Issue #24, opened 2012-02-24
Write some fucking tests


OscarGodson:
Gots to be done. Will probably use @foobarfighter 's https://github.com/foobarfighter/foounit

OscarGodson:
HOLY SHIT I CAN FINALLY CLOSE THIS! 55 passing tests later, we should have full coverage of every method and event

![tests](https://img.skitch.com/20120429-t973yi41e7snm22f8nybt3ajq.png)

Issue #21, opened 2012-02-21
Support data-N attributes for all settings


OscarGodson:
Allow settings to be set via `data-N` attributes. Example:

``` html
<div id="my-editor" data-ee-focus-on-load="true" data-ee-default-content="hello world"></div>
```

OscarGodson:
This is basically an expanded version of ticket #20. Closing #20 in favor of an all encompassing solution.

Issue #20, opened 2012-02-21
Support data-filename


OscarGodson:
You should be able to open a file via the `data-` attribute like:

``` html
<div id="my-editor" data-filename="foo"></div>
```

Currently it'll make a file called "my-editor" if no different `.open()` is used.

OscarGodson:
Dupe of #21.

PR #38, opened 2012-03-05
Experimental api update and test suite

johnmdonahue:
RE: [Issue 24: Write some fucking tests](https://github.com/OscarGodson/EpicEditor/issues/24)

I started to add some tests using `foounit` and in doing this it seemed like a handy option to configure the localStorage name. I refactored the code to remove the `options()` method and instead allow the EpicEditor constructor to simply accept an options object that includes the target element. I also added 4 simple tests to just get things moving.

``` js
var opts = {}
  , editor = new EpicEditor(opts)
```

I considered trying to work with the current options method, but the localStorage is set from within the constructor and I couldn't think of an application where changing this option could be useful after initialization.

This could just as well be accomplished by `EpicEditor(el, opts)` or extracting the localStorage definition to the existing `EpicEditor.options()` method.

Let me know your thoughts on this. Didn't want to get too far before checking in.

Also what are your thought on setting up [node-jshint](https://github.com/jshint/node-jshint) for contributions?

Love the tool and looking forward to contributing however I can.

Thanks.

John

PS: Don't really know of a better way than a pull-request to share this but it's def not pull-request ready. More just a conversation starter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ EpicEditor is an embeddable JavaScript [Markdown](http://daringfireball.net/proj
 WYSIWYGs suck and they suck hard. Markdown is quickly becoming the replacement. [GitHub](http://github.com), [Stackoverflow](http://stackoverflow.com), and even blogging apps like [Posterous](http://posterous.com) support Markdown now. This allows you to generate a Markdown editor with a preview, fullscreen editing, full CSS theming, and offline support with a simple:
 
 ```javascript
-var editor = new EpicEditor(element).load();
+var editor = new EpicEditor(options).load();
 ```
 
 ###How
@@ -37,8 +37,8 @@ $ git clone git@github.com:OscarGodson/EpicEditor
 #####Step 3: Init EpicEditor
 
 ```javascript
-var element = document.getElementById('editor-wrapper');
-var editor = new EpicEditor(element).load();
+var el = document.getElementById('editor-wrapper');
+var editor = new EpicEditor({element: el}).load();
 ```
 
 ####API
@@ -52,7 +52,7 @@ The constructor is first (`EpicEditor()`), but everything after are methods of t
   <li><a href="#api-epiceditor"><code>EpicEditor()</code></a></li>
   <li><a href="#api-load"><code>load()</code></a></li>
   <li><a href="#api-unload"><code>unload()</code></a></li>
-  <li><a href="#api-options"><code>options()</code></a></li>
+  <li><a href="#api-options"><code>Customizable options</code></a></li>
   <li><a href="#api-get"><code>get()</code></a></li>
   <li><a href="#api-open"><code>open()</code></a></li>
   <li><a href="#api-import"><code>import()</code></a></li>
@@ -76,7 +76,7 @@ Creates a new EpicEditor instance. Give it an element you want to insert the edi
 **Example:**
 
 ```javascript
-var editor = new EpicEditor(element);
+var editor = new EpicEditor(options);
 ```
 
 _Note:_ all the examples below will continue to use this same constructor.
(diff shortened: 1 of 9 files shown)


OscarGodson:
I was actually thinking about the exact same options thing today. When i first started the project the idea was that you could `unload()`, change options then call `load()` again or just call a `reload()`. But reloading the entire thing for a single option change would be needlessly intensive. @dadambickford for example wanted a way to swap out themes on the fly, so instead maybe add a parameter to `.preview()` that allows you to pick a theme.

So, enough back story :)

I _love_ your idea. I'd prefer if options was a second optional param since the element param would always be required. So it'd be `new EpicEditor(el,opt)`. What do you think? I feel like thats the most intuitive and clean method.

Yeah, adding jshint would be great. I'd like to get a deploy process where once we have a master commit it runs tests, runs JSHint, then runs through the Closure Compiler API (http://closure-compiler.appspot.com) and if all is good, push the commit. 

I'll clone your repo and go to your branch and checkout your tests, but it looks really awesome. Way more than I was ever expecting! I'm thinking if you get a decent amount of tests written and get the options thing refactored we could merge this into the "official" beta commit since your options change will be a breaking update.

johnmdonahue:
This sounds great. For reference, my thought on having a single opts param was to be to able to simplify with a default element like this:

``` html
<div id="epiceditor"></div>
<script>
</script>
```

In the easiest case nothing would need to be passed unless the user wants to use a different el or define other opts. Also, it might actually be cleaner to pass the el id rather than its object:

``` html
<div id="mySuperCustomEpicEditor"></div>
<script>
  var opts = {
      targetElement: 'mySuperCustomEpicEditor' // rather than $('#mySuperCustomEpicEditor')
    }
    , editor = new EpicEditor(opts).load();
</script>
```

That said, I don't really have a strong objection to the `(el, opts)` route either, so I'm down to work on it however.

I'm happy to jump on both the options refactor as well as core API test suite. Just let me know any additional feedback.

Thanks!

OscarGodson:
Never thought about having a default id, but i like that. I actually wanted to be able to set options directly in the HTML with namespaced `data-epiceditor-*` attributes (see: #20 and #21). With your method you could potentially have a single call somewhere and never have to touch the init code again, just swap out the HTML based on user preferences or something.

Actually @techwraith wants to add this to his Scotch project. I'd like to get his opinion on this too since he's one of the first people who'd be using it in a real world blogging CMS. I'm actually up in the air about it now since i see your point about having an id that it automatically looks for. I'll have to think about this for a bit.

You could also potentially still have `new EpicEditor(el,opts)`, but have `el` be an id of an element. Just another thought.

johnmdonahue:
+1 ... I dig this, `data-epiceditor-*`

``` html
<div id="my-editor" data-ee-filename="foo" data-ee-storage-name="bar" data-ee-editor-theme="dark" data-ee-preview-theme="github" data-ee-focus-on-load="true" data-ee-default-content="hello world"></div>
```

dadambickford:
Since I'm still very much a student of JS, I was first thinking, "noooo don't change it! I just got my head wrapped around it!" but I started to comprehend it while I was driving today, and I like having the option object passed as a parameter instead of the `.options()` method. haven't had much exp with using `data-` attributes but that looks like it could be convenient.

johnmdonahue:
I have a little time this week I can work on this but am headed out of town for a long weekend :-). I'll wait for @Techwraith to weigh in before jumping back on it. Just let me know.

OscarGodson:
I talked with him in person today. We liked the idea of EpicEditor(opt) with no selector. So do what you had originally of using the ID epiceditor. Great idea :)

johnmdonahue:
Sweet. I'll try to make some headway this week. Thanks.

OscarGodson:
One tiny thing, can we use the default /spec directory for tests?

johnmdonahue:
No objections here, I understand `/spec` from a rails convention but I've just become accustomed to seeing `/test` in the javascript/node arena:

https://github.com/joyent/node/
https://github.com/jquery/jquery
https://github.com/visionmedia/express
https://github.com/flatiron/flatiron
https://github.com/ender-js/Ender
https://github.com/twitter/hogan.js

etc. All use `/test`. I see that `foounit` is using `/spec` but it didn't recommend or create that dir so I just went with `/test`:

> Create a directory to put your tests

OscarGodson:
It's the default for foounit, but very good point, lets go with /test then :)

OscarGodson:
This is basically #58 so closing

PR #42, opened 2012-03-06
Minor Change

jwmcpeak:
Just a minor change, but changes _getStyle() to test for standards compliance first--important for browsers (IE9+) that support the standard API while also supporting legacy/proprietary APIs. Also defines y variable once as opposed to twice.

### DIFF
--- a/epiceditor.js
+++ b/epiceditor.js
@@ -60,13 +60,15 @@
    * @returns {string} Returns a string of the value. If property is not set it will return a blank string
    */
   function _getStyle(el,styleProp){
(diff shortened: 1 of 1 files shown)


OscarGodson:
Awesome, much thanks. Ill take a peak at this later today and make sure everything works as expected manually since im still setting up tests with @johnmdonahue [#38] so I can't have it run through a test to check, yet.

dadambickford:
verified working in IE9, ff, and chrome. Merging in per Oscar's request!

PR #53, opened 2012-03-12
Minor changes

jwmcpeak:
Added radix to some parseInt() and parseFloat() calls to prevent erroneous parsing; modified _insertCSSLink() to not set media property twice; moved isFunction() out of _mergeObjs() and renamed to _isFunction();

### DIFF
--- a/epiceditor.js
+++ b/epiceditor.js
@@ -78,11 +78,11 @@
    * @returns {int}
    */
   function _outerWidth(el){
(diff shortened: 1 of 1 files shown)


OscarGodson:
Nice, great additions and good catch on setting `var` multiple times and the CSSLink stuff. Will test and pull it in! Thanks for these pull requests :)

PR #58, opened 2012-03-18
0.1.0 build refactor

johnmdonahue:
Notes:
- `FooUnit` is not yet integrated but I wanted to get your thoughts on the current format.
- `Jakefile` works but probs needs some love
- I originally had `onevar` set to true in the JSHint config, and I personally like this as it seems to keep code cleaner and forces me to break out into smaller functions when things get hairy. However, I know the fullscreen code is being worked on right now and this section has several `var`'s scattered throughout. I'm guessing some of these functions will be pulled off `load()` out with the other helper utils?
- I tried to only make trivial changes in getting the code to pass jshint, nothing too crazy. I left a couple `@NOTES` in the comments.
- I made another update to the API, and please feel free to reject this change - but `import` like `export` is a reserved word. Maybe not that big of a deal but I figured for consistency with `exportHTML`, `importMarkdown` made sense anyways.

Let me know your thoughts on this. Thanks.

Hacky Saint Paddy's Day!


OscarGodson:
For the var stuff, great! Yes, like the `_saveStyleState()` for example and a lot of that is probably gonna get moved around and outside `load()`. I noticed somethings that you fixed that @jwmcpeak actually fixed too (like where I accidentally set var multiple times on some helper functions), so it might be worthwhile to merge in master if it's possible or check his diffs that had some fixes of this sort.

I think your changes are fine to make it pass. You made a minor change to the `_mergeObjs()` which _should_ be fine, but that's actually using jQuery's `$.extend()` method so not entirely sure if there was a reasoning behind leaving it off, but it should be fine.

For the `import()` stuff, good catch. I like `importMarkdown()` for myself, but I wanted to make the editor as easily extendable as possible so you could potentially use it for a Wiki editor, or (god forbid) BBCode, etc. Just drop in a parser/lexar, make a tiny change to `exportHTML` and it'll work. Maybe we swap the word `Markdown` in the method name to be more ambiguous like `importContent()`?

Now... gonna pull it down and play with it :)

johnmdonahue:
Cool, maybe we can just leave `onevar` on with failing tests until we work through this iteration. (commit pushed)

From what I can tell, the changes you mention are already on the initial 0.1.0 branch. I see in the last merge that `_isFunction()` is pulled out as a private method which makes sense, this could be used in `load()` and `unload()` where the callback is checked. Also it looks like another possibly unused line was deleted. But I also saw that a radix was added to `parseFloat`, which isn't necessary. All commits from master appear to already be pulled in here, but I am pushing an update to remove the radix on `parseFloat` here.

As for `_mergeObjs`, I'm not exactly sure what you mean unless you're referring to the `hasOwnProps` addition, everything else was basically whitespace changes.

I understand `importContent` as a name, but with unspecified syntax would you then want to require a type be passed in to let the parser know what to expect? Are you proposing that someone should be able to import textile for instance and override the renderer? If so, it'd be good to have this in mind when thinking about structure. This could be cool, I just hadn't thought about this way. 

Thanks.

OscarGodson:
Yeah, I was just talking about the `hasOwnProperty` check. Since it was taken from jQuery, wasn't sure if there was some reasoning since so many people have touched the code.

I was talking about #53 or sha: c1de1bfd65e714a3cecca071b077a86b85eea984 (he also took out the `var` in `_outerWidth()` and `_outerHeight()` is all).

Good idea about importContent and adding a 2nd param. It's something I'd like to add on later down the line is be able to allow filetypes and this is also why I tried to leave this part of the code so "blackbox" like. If you have other name suggestions or just other ideas altogether let me know. The goal, later down the line, was a way for developers to add parser/lexars like:

``` javascript
EpicEditor.extend({
  name:'wiki'
, type:'parser'
},function(){
  //parser code...
  return html;
});
```

Then do `importContent('content...','wiki');` or something. Would like to get your thoughts tho.

If you think it's a good idea it might also make sense to think about it for a near release to eat our own dogfood and use a `.extend()` like method in our own core code for the MD. I dunno... just thinking out loud.

OscarGodson:
The pull looks pretty good, except two things:
1. In the index.html file you link to `/epiceditor/...` it should probably be `epiceditor/` so that it doesn't break when it's not cloned into the root of your server.
2. Not sure why, but `jake build` isn't writing anything. In fact, doing it makes the file blank. The `buffMin` `var` also seems to be blank. I haven't played with it a lot but not sure why.

Otherwise, it's lookin sexy ;)

johnmdonahue:
1. Good catch on the `index.html` file. Updated.
2. Hmmm. Strange it's working fine on my end. Do you have `uglify-js` and `JSHint` packages installed globally? That said it'd probably make sense to update the `readme` with documentation on building/linting using the `Jakefile`. And also provide errors when either tool is not found. Definitely needs some love.
   
   npm install -g uglify-js
   npm install -g jshint

OscarGodson:
Ah, that's probably it, duh. I figured I'd get an error if it didn't
exist tho. I'll try it again today.

On Mar 19, 2012, at 10:35 AM, John Donahue
reply@reply.github.com
wrote:

johnmdonahue:
Ha, yeah at a minimum there should be instructions and error checking here. Like I said, it needs some love but it could be a workable starting point anyway.

johnmdonahue:
I pushed the original tests. They definitely need more work, but it's a start.

I ran into a few scoping issues with foounit in the process, including [this one](https://github.com/foobarfighter/foounit/issues/32) which led to [this modification to foounit code](https://github.com/johnmdonahue/EpicEditor/commit/bd702a9a13579afaab3db04f062f139873d6ebc0). I also needed to wrap Marked in the main EE wrapper to prevent other scoping conflicts with foounit's wrapper which then meant the `jake lint` task needed to be updated to concat before lint. 

It's pretty duct taped together at the moment, but hopefully a workable start :)

OscarGodson:
Superb! Next time you can just send me a pull request to my `0.1.0-build-refactor` branch if you want.

If that's a bug with foounit you should @foobarfighter a pull request.

Thanks! i'll see about pulling this into the `0.1.0-build-refactor` branch and start getting some tests written :)

johnmdonahue:
It looks like this was just automatically added to the pull request as you mentioned [here](https://gist.github.com/74f99b5585d8631d08ba#gistcomment-194723). Next time I can target your branch if you prefer. 

As for the foounit scoping issue, I submiited this bug report: https://github.com/foobarfighter/foounit/issues/32, but didn't really have time to explore it much further. It's probably more than just a simple fix - maybe more of a methodology issue. My fix was just a workaround. Either way, I wanted to get foobarfighter's opinion before jumping in the code although I'd be happy to.

OscarGodson:
I see. No biggie.

And I'll ask him today if he comes in too.

On Mar 27, 2012, at 10:23 AM, John Donahue
reply@reply.github.com
wrote:

OscarGodson:
So, in person he told me:

> I think module is part of the commonjs spec, so i dont know if this is really a bug

I havent gotten time to really look at the bug yet, but tonight I can.

johnmdonahue:
Yeah, it was late and I'm sure I'm missing something here. It may just be the way we have things set up causing the problem, or Marked relying on `module` to check. Not sure if it's a foounit issue at all. ;)

For reference, though, the issue boils down to that in a browser context `module` is not defined but in Node it is:

``` bash
$ node
> typeof module
'object'
> 
```

``` javascript
typeof module // "undefined"
```

OscarGodson:
This has actually been merged into EpicEditor's official repo under the same branch name (`0.1.0-build-refactor`) and @johnmdonahue is contributing directly to the core, so closing this.

PR #70, opened 2012-03-27
Fixes undefined error for optional callback.

chrisnicola:


OscarGodson:
When would this happen?  `callback` should always be at least `function(){}` from the `callback = callback || function () {};` line.

chrisnicola:
There is no such line in the unload function, but you are correct I could have done that to match the load function which does have that.

OscarGodson:
@lucisferre My bad, yes, in master this isn't there. @johnmdonahue added this in 3531be85 on the `0.1.0` branch. The `0.1.0` branch is a WIP, but is fairly stable as long as you dont need IE9 for now.

Also note that if you're serious about using EE for anything I'd work with `0.1.0-build-refactor` branch as that'll be our final branch which has breaking API changes, actual builds as well as tests.

https://github.com/OscarGodson/EpicEditor/blob/0.1.0/epiceditor.js#L704

Thanks a lot for sending the pull request and I'll make sure to add tests for load and unload's callbacks now :)

chrisnicola:
Thanks Oscar, I'll switch to that soon then.

PR #91, opened 2012-05-07
Jakefile Cleanup

johnmdonahue:
Here is the first draft at a cleaned up Jakefile.
### Features
- Hopefully able to run on Windows since no more reliance on `cat` / other unix calls
- Added nolint flag option for build task
- Cleaner simplified logic and code
- Using new Jake API. Good shit btw!
- Fixed broken color val for green making it show up magenta
- Trimmed down some extraneous output, just LMK if you want any of it back in
### ToDo
- Check for `jshint`, `marked`, and `uglifyjs` (or otherwise display a warning)
- Do we need pretty errors on jake.exec? Not sure how to implement this ATM
- Use jake package task to generate zips in `docs/build`?
- Figure out a good way to version the builds in the header etc.
### Notes
- `jake.FileList()` is not being used because it requires `minimatch` - trying to keep it simple
- What's the best dir structure to handle the build and docs? `index.html` and builds moved into docs? Create another folder `/build`? Something just seems odd about having a folder called docs and the index in the root - can't really put my finger on it.
- If you have a node windows box kicking around I'd love to hear how bad this shit crashes it. :P


johnmdonahue:
I'm sure I've probably missed some things we discussed so if you can think of any just tack them on here and I'll try to get to it ASAP.

OscarGodson:
@dadambickford is on Windows ATM. @dadambickford could you check out and post comments about Windows issues here?

I dont think we need pretty errors. As long as dev can copy and paste the error blob.

Yeah, a zip would be cool. Are you thinking per build or per release? I.e. a `-z` like command to zip the build before pushing a tagged commit.

For the version thing in the docs we should just have a way to get the version in EpicEditor. Maybe like `epiceditor.epiceditor` or `epiceditor.version`. jQuery does this and it was actually really handy when we were converting from 1.4.3 to 1.7.1. (`jQuery.fn.jquery;`). Also handy if you wanna force a cache refresh (i.e. the schema stuff is changing from 0.0.3 to 0.1.0 so I can do like `if(oldVersion){ localStorage['epiceditor'].clear() }`

Oh, and I like /build, but we will still need the `index.html` file in the root for `gh-page`'s and stuff unless you wanna make a separate build for `gh-page`s but that seems like a pain in the ass.

johnmdonahue:
As for the zip, yeah Jake has a cool new `PackageTask` I was wanting to play around with. Basically I was thinking it would be nice to be able to download the built `epiceditor` folder directly from the docs similar to the way [GH Project Pages](http://pages.github.com/) does or Bootstrap. A lot of front end devs might like the option to just download the latest build rather than deal with the repo directly.

RE: https://github.com/OscarGodson/EpicEditor/pull/93#issuecomment-5561531

Yeah. There may be some kinks to work out but like you said it's a dev branch anyway. I'm going to go ahead and merge. I feel like it's probably safer than the existing one anyway. ;) Just easier to keep track with so many branches floating around.

@dadambickford let me know if you run into any issues or have any questions.

Thanks

OscarGodson:
Sweet! one branch down and I'll be able to play with this tonight some.

PR #105, opened 2012-05-11
Added Twitter link to the auto links section

roberocity:
Will convert `@username to` `<a target="blank" href="http://twitter.com/username">@username</a>` just like it works for email addresses and other basic URL patterns.


OscarGodson:
Hmm, this is interesting. We're moving to @chjj 's Marked in the very soon to be released beta because showdown is terrible, but I kinda like this idea. @johnmdonahue How do you feel about this but doing it in Marked somehow?

johnmdonahue:
I like this idea, but maybe as an extension rather than core. I am thinking of situations where this may cause unexpected outcomes for users not intending to point to Twitter. Say I want to write the following:

``` markdown
## Contact
My handle is @foomanchu pretty much everywhere except on twitter where I'm @foobazzchu.
```

This seems like a perfect candidate for an extension but `@links` in general can refer to GitHub `@users` and other apps that use similar conventions. And those who don't care about any of that might get confused about why `Get @me dog!` gets converted to a Twitter link.

johnmdonahue:
BTW. It looks like a similar request was brought up in Marked as well: https://github.com/chjj/marked/issues/44

chjj:
@johnmdonahue, just to let everyone know, I leave a lot of issues open because I _want_ to be challenged on my opinion. Even if it looks like I've decided on an issue, I probably haven't. I just have strong opinions on it.

roberocity:
@johnmdonahue I'm a fan these type of rules being extensions rather than part of the core. Keep the core clean and simple and provide a point to add additional rules. The point about the `@` being common for other sites is another reason to have extensions. Conflict would ensue either way and avoiding it may be difficult. 

In the grand scheme I think that twitter (almost, perhaps) owns the `@username` and if other `@username` were needed they could use the full link or an escape character perhaps to avoid creating the link `\@username` to prevent the link.

Still an extension is called for and not part of the core syntax.

roberocity:
@OscarGodson do you plan on having Marked as a client-side requirement or will you bundle it together with EpicEditor similar to how you have with Showdown?

OscarGodson:
I like the idea of the extension. The extension would allow you to pick the domain to add it to as well, something like:

```
EpicEditor({
  mentionizer: {
    basePath: 'http://twitter.com'
  }
});
```

I'm going to keep this pull request open as a stub for this extension.

OscarGodson:
@roberocity Once the extension stuff is done it's going to be pulled out and will be added as a "parser extension". For the beta coming (0.1.x) it'll be bundled. Or, this is what we've had planned.

OscarGodson:
Now there's a parser option so you could do this in that.

ryanramage:
I have stated a markdown flavour with a mix of markdown and twitter markup here:

https://github.com/eckoit/bird-down

I am currently using it with epic editor successfully, using the custom parser, see here for an example usage:

https://github.com/eckoit/my-path/blob/72efb1ceb27ebb8a4f95e80344a68d84f7ed6eeb/js/quick.js#L74

OscarGodson:
Awesome! Always glad to see what people are doing with EpicEditor :)

OscarGodson:
Oh also, if you haven't already, add your script to the EpicEditor Implementations page in the wiki!

PR #108, opened 2012-05-16
Nicer download btn and section

OscarGodson:
Dear @johnmdonahue,

Added a nicer download button and moved it to a new 'Download' section which in the future could contain multiple versions like stable, beta, and alphas.

The only issue is that I added this to the README.md so that we could have the download section also in the README. You didn't wanna update more than one place which I totally understand. What if we we use `@VERSION` in the README also? Would that work? Then if that works we can swap the `zipUrl` var with simply the `href` of that download link and then everywhere is updated even in the README.

I also moved around some of the main.js so that when we add sections it doesn't break stuff by using the IDs of the elements rather than counting them.

Love,
Oscar :heart:


johnmdonahue:
Dearest @OscarGodson,

Love the button!! And great idea on the main.js tweaks. Counting was just a hacky way of getting the concept rolling - happy to see a more reliable implementation. 

A few thoughts on the other points brought up here. My main concern is to keep things as simple as possible, both for the user and the contributor:
1. I don't think the `.zip` shares the same relative path on GitHub as on the docs, so `docs/downloads/EpicEditor-v0.1.0.zip` wouldn't work - it would be located here no?: https://github.com/OscarGodson/EpicEditor/blob/feature/download-btn/docs/downloads/EpicEditor-v0.1.0.zip - or similar on master - and that won't directly download anyway (this is actually something I have been wanting to address for the logo as well). Maybe we could just link to the docs from the `README` - e.g. [Download the latest stable release here.](http://epiceditor.com) and then swap out for the fancier button with version number on the site. I don't think I've seen any other projects on GitHub provide direct download links from READMEs and this makes sense to me. It's reasonable to assume that users who have found their way to GitHub either know how to use it or can find their way to the docs provided obvious links are available.
2. Using `@VERSION` in the README would require a separate source `README.template` to build the `README` and this seems like an unnecessary addition. Unless you were thinking of some other way?
3. In my opinion, multiple versioned downloads could be distracting and tough to maintain. [jQuery](http://jquery.com/), [Backbone.js](http://documentcloud.github.com/backbone/) and [Twitter Bootstrap](http://twitter.github.com/bootstrap/), for example, all only provide direct download buttons to the latest stable release. People wishing to dive deeper, contribute, or play with beta features can do so from GitHub. I think it makes sense to only publish one build of the project for any given tagged release and I believe this is how jake.PackageTask is set up to work - i.e. it will overwrite the package dir with only the current build.

Let me know your thought on this.

Sincerely,

John :cake:

johnmdonahue:
Maybe we should style the try it button to be more like the download too! Also I noticed that it now injects the try it demo in the navbar, which actually looks kind of cool, but I'm guessing was unintended. :P

OscarGodson:
1) Makes sense. How does the GitHub downloads tab work? I see other projects have all their versions in their as zip, but I'm not entirely sure how that works. A little confused on what you mean by:

>  swap out for the fancier button with version number on the site.

Do you mean have the button with version number just become a "Download" button that is actually just a link? The Download section I think is still needed whether it's a link, download, or whatever tho so we have some call to action on the site. Also, It's probably good to have "Download" in the left hand nav as well.

2) Duh, the README would literally say `@VERSION`. Darn.

3) Sounds good. I was just thinking we _could_.

Automation is hard.

OscarGodson:
@ your 2nd comment:

**oops**

haha

johnmdonahue:
I think the downloads tab is to download the entire repo as a zip, not a build, but I could be wrong.

RE: Swapping out the fancier button - basically just replace the "Download the latest version here." on the README with the new button html and link via jQuery on the docs.

OscarGodson:
I dunno. The download tab to me means a download section since all the other links are sections. If you thought it was a download you'd just be redirected to that section where you could click the download. We can just skip that as a left hand nav item tho like I did with the "An embeddable..." headline by just incrementing the number by 1. (we can change this to `2` https://github.com/OscarGodson/EpicEditor/commit/4bc5387b2f17f6d30193b45ca6042b591fab5700#L3R17)

If you have other design suggestions tho feel free mod the branch and push it and let me check it out.

johnmdonahue:
>  How does the GitHub downloads tab work?
> ...
> The download tab to me means a download section since all the other links are sections.

I'm a little confused. Are you talking about a downloads tab like this: https://github.com/jquery/jquery/downloads or https://github.com/jquery/jquery/tags ?

As for the download link in the TOC, yeah, it makes sense to me to remove it. And maybe even just by removing its header. I think the button is kick ass enough to speak for itself without a heading. :)

I'll play around with this some tonight.

johnmdonahue:
Ahhh! I was replying to this:

> How does the GitHub downloads tab work?
> I think the downloads tab is to download the entire repo as a zip, not a build, but I could be wrong.

Not this:

> Do you mean have the button with version number just become a "Download" button that is actually just a link? The Download section I think is still needed whether it's a link, download, or whatever tho so we have some call to action on the site. Also, It's probably good to have "Download" in the left hand nav as well.

OscarGodson:
The first, the https://github.com/OscarGodson/EpicEditor/downloads

Do you actually have to manually upload shit? If so that's a pain, or is there like a github hook you know of?

johnmdonahue:
Don't know of any hooks for that. It seems like the only automated downloads are for full repos on the tags and downloads tabs. I've never used that to grab projects anyway, have you? Seems like it was intended for binaries and things that are compiled from source but not necessarily found in the source repo. My guess is that the majority of users will hit the docs (index.html) download button primarily or just clone the repo, so need to mess around with uploading builds there.

johnmdonahue:
Pushed a few changes. Let me know if these help at all. I think things are looking pretty sweet myself. I slapped that button style on the try it button too!

Let me know if you want to figure out a way to handle multiple version downloads. I think I will probably need to rewrite the way the package task is handled.

Thanks.

OscarGodson:
I'm fine with one download. We can figure that out later (if ever) if there's a situation when we need it.

Like the try it button _a lot_ better now. Was so ugly before haha.

If you approve, click the merge button :wink:

OscarGodson:
Also made this very important ticket (#110)

johnmdonahue:
Yup. Looks like `body` has `text-shadow` set and when I lightened the background color on `pre` it became noticeable there. I just pushed a fix setting `pre` to `text-shadow:none`. Cool by me. I'll merge and close unless there's anything else?

johnmdonahue:
Actually I'm off to bed. Merge away!!

:poop:

OscarGodson:
AWH YEAAAAAH! The indentation was totally screwed up for me in vim where @SEBNITU put tabs in :tongue: but I was too lazy to fix at the time haha.

PR #114, opened 2012-05-21
Parser as option

johnmdonahue:
A bunch of updates here so let me break it down. Also I realize I didn't get much done on tests specifically, but the good news is, it now passes the linter :) ... and I'll be cleaning that up more this week.
1. **Added linting for other JS:** Now `jake lint` will check all of the JS files listed below. `jake build`, `jake docs`, and `jake test` only lint their associated files. I'm not sure if you would want a force option for each so that is currently only available for `build`.
   - `src/editor.js`
   - `Jakefile.js` (Note: `jshint` does not currently support files without ext so .js was added.)
   - `.jshintrc` (just because it's funny to lint a lint config)
   - `docs/js/main.js`
   - `spec/spec.js`
2. **Pulled Marked out of EE wrapper:** The intro.js and tmp file concat is no longer necessary. The default EE build will still use Marked but an optional parser can be provided by the user as described in 3. The only real drawback of this is that it creates an additional global but this seems like a reasonable thing to accept in trade for the flexibility it offers.
3. **Added settings.parser:** This option accepts any function that returns a string. If the user has included `Showdown` or some textile parser on the page, the parsing function can be passed into this option otherwise EE will look for Marked by default and if that is not available, use a generic function that returns an error message followed by the original content wrapped in `<pre>` tags (See example below.). Putting the parser on the settings object seems like a great option rather than extending the EE object as we had discussed. I am imagining an implementation where a user is allowed to select between textile or markdown to parse their content. As a part of the settings object, this parser could be toggled at any point for a given instance and still function (e.g. realize your content is textile?... just switch the parser).
4. **Added build option to use custom parser:** Running `jake build parser=path/to/parser.js` will override the default Marked build and replace it with the custom script. You can play around with this as follows:

Clone Showdown:

``` bash
$ git clone https://github.com/coreyti/showdown.git
```

Build EE using Showdown:

``` bash
$ jake build parser=path/to/showdown/src/showdown.js
```

Modify `docs/js/main.js` to point to the Showdown parser:

``` javascript
  var opts =
      { container: 'example-1'
      , file: { defaultContent: "#EpicEditor\nThis is some default content. Go ahead, _change me_. " }
      , focusOnLoad: true
      , parser: new Showdown.converter()
      }
```

Whoops! That's not the rendering function.

Now correct the parser:

``` javascript
  var opts =
    { container: 'example-1'
    , file: { defaultContent: "#EpicEditor\nThis is some default content. Go ahead, _change me_. " }
    , focusOnLoad: true
    , parser: new Showdown.converter().makeHtml
    }
```

Good to go.
## Current Jake tasks

``` bash
jake lint           # Lint all js files  
jake lint:editor    # Lint core EpicEditor: src/editor.js  
jake lint:docs      # Lint doc related js: docs/js/main.js  
jake lint:spec      # Lint test related js: spec/spec.js  
jake lint:util      # Lint utility and config js files  
jake build          # Build epiceditor.js and minify to epiceditor.min.js  
jake build:force    # Force build epiceditor.js and epiceditor.min.js skipping pre-reqs  
jake docs           # Build index.html from the README  
jake test           # Test code against specs  
jake package        # Build the package for distribution  
jake clobber        # Remove the package  
jake ascii          # Kick out some ascii
```


OscarGodson:
Probably don't need a force on all of those. Force is just handy for debugging and we'll probably not be doing much hardcore debugging in the docs.

As for 2, it was already global. Marked was adding it to the window (this is how I was debugging it earlier)

3 and 4 cool, love it. Never thought of that, but great idea and a lot easier. We'll need a way to update settings on the fly tho if you want to be able to update the parser programmatically which we'll want to have. I believe I have a ticket for that already for updating themes on the fly.

OscarGodson:
Oh, and i'll play with this maybe tonight (gotta work on actual work stuff tonight) and pull it in if nothing breaks for me either.

And, we should add all these jake tasks to the contribution guide or at least tell them about `--tasks`.

johnmdonahue:
Yeah, on 2 that's true. I thought we had discussed trying mask the parser in future builds, but I don't think we need to anyway. As for toggling the editor it should be as easy as changing the setting. It just needs the pointer to the function, right? So my thougts were that even after initialization you could do something like `editor.settings.parser = newParser` and maybe refresh the preview if already in that mode.

OscarGodson:
Awh, yes. that'd work. Probably wanna document that you can do `.settings` since right now it's undocumented (maybe in the `options` section?), but you could do that. An easy example would be:

``` javascript
editor.settings.parser = newParser;
editor.preview();
```

OscarGodson:
Looked at the updates and don't seem to see anything noticeably different but I noticed the tests take about 10 seconds now rather than 8 for me. Any real changes other than aesthetics? Hard to tell since so many lines were updates ;) 

Not a big deal, but I figured you made or added something to make it take 2 more seconds.

johnmdonahue:
No code changes apart from white space but test optimization is what started this mission and it's next on my list.

OscarGodson:
Different pull request or the same one? Do you want me to just wait for the optimization stuff?

OscarGodson:
Haven't been able to take a look really, but is this from you?

 

Seems very strange... not sure how that got in there.

johnmdonahue:
I'll clean up the tests separately. As for "doing interval stuff" I have no idea where that's coming from. Where are you seeing it? Local storage? I don't see anything like that on tests or otherwise and it doesn't seem to be in the code anywhere. Very strange. That said, it does seem like something I would do ;) ... Could it be from an old branch?

OscarGodson:
Very well could be. Also, @dadambickford is getting slow tests like you. He's on Linux/Ubuntu Chrome and getting 33 seconds to run the 67 tests. I wonder whats going on. Its either fast like 8 seconds or over 30 seconds. No in the middle. Hmm.

johnmdonahue:
Yeah. I just ran tests again looking for "interval stuff" and got `>> foounit runtime: 55583ms`. I have some ideas on speeding it up that I'll try out this week but I'll work on that separate from this pull request. Anything you want to add/subtract here or are you down to merge? Ideally I'd like to work on the tests from this point since `spec` now passes lint.

OscarGodson:
Done!

johnmdonahue:
Awesome! How do you feel about renaming/merging `0.1.0-build-refactor` to `0.1.0` and removing merged/extraneous branches? Feel free to remove any branches I've thrown out there.

``` bash
$ git branch --merged 0.1.0-build-refactor 
* 0.1.0-build-refactor
  feature/crud
  feature/download-btn
  jakefile-cleanup
  parser-as-option
```

OscarGodson:
I'd like to rename it also. It's been bugging me as well. Those all look OK
to delete also. You can delete them if you want or I can tonight or
sometime this week.

On Mon, May 21, 2012 at 1:41 PM, John Donahue <
reply@reply.github.com

OscarGodson:
@johnmdonahue we might not want to rename the branch actually now that I think of it. AnchorCMS is looking into it and so is Punch (http://laktek.github.com/punch/) now too and both on that branch and those are just two I've thought of. Maybe we should just hold off on that since the beta is so close anyway.

johnmdonahue:
@OscarGodson ahh yeah. Good call. We should be ok to delete merged topic branches though right?

OscarGodson:
Yep!

On Tue, May 22, 2012 at 11:05 AM, John Donahue <
reply@reply.github.com

OscarGodson:
After looking at this, should we have HTML spit out for the error or just throw an error? I like the idea of just throwing an error like:

error

johnmdonahue:
Yeah, that makes sense to me. Are you thinking we should still provide a basic "parser" that just wraps the input in pre tags but remove the error message? Or disable the preview button? How should we let the end user know that something has gone wrong or is this even EE's responsibility? 

Maybe we could get around all of this by simply providing an error event: `ee.on('error', handleError)`. If we could provide a hook then developers can be as thorough as they like in handling errors. For instance, some other script kills the custom parser, an error event is fired and the developer might attempt to reload the parser... just thinking out loud here. Not sure if this is overcomplicating things.

OscarGodson:
An error event would be a great idea. I'll just put that into this commit. I don't personally think it's EE's responsibility. The only time it'd fails is:
1. Parser isn't a function (shit just breaks)
2. Parser doesn't return a string (you'll get something like `[object][Object]` or `undefined` in the previewer)
3. Parser get's clobbered like you said from an external source. (same as 1 probably)

and all of those will be pretty clear that the previewer is busted and with your idea of `on('error')` we can give them the error to do what they please. This could be injecting the previewer with what you had. It'd be as simple as:

``` javascript
ee.on('error', function(data, err) {
  if (ee.message.indexOf('parser') > -1) {
    ee.getElement('previewer').body.innerHTML('ERROR: ' + err.message);
  }
});
```

or something...

OscarGodson:
Do you think i should even throw errors then or simply emit error events?

johnmdonahue:
Yeah, I like the idea of emitting _rather_ than throwing errors. That way the user can just hook into the event and not have to worry about catching errors.

OscarGodson:
sounds good!

johnmdonahue:
RE: https://twitter.com/pamelafox/status/202103185576443904

Hey man, I saved this tweet a while back thinking maybe there was something to it and now I see it... What Pamela Fox seems to be looking for would be super trivial to set up with EE by just disabling the parser.  You could do straight-up full-screen html editting in real time :D



I can imagine cases where people are just wanting to mock something up or maybe create a tool that teaches people to use HTML etc. I'm not sure how this fits into the EE as a markdown editor concept, but we've already talked about being able to use it for Textile etc. EE can really be any kind of editor as long as there's a parser that outputs HTML. You could even hijack the preview pane and turn it into a "console" if you wanted.

Anyway, just wanted to throw that out there. If you're interested, I can push the change that makes EE just output the input if no renderer is found or fails to return a string - effectively turning it into a real time html editor.

OscarGodson:
Yeah, that'd be rad! :heart_decoration:

If it fails to return a string that would effectively just break it tho, no? What are you going to do if it fails to return a string? Are you gonna take the innerHTML of the editor? I like the idea of it defaulting to just use plain ol text (which in turn does what you're showing above), but not sure about "guessing" what the dev wants. Could we still have it fail if it returns an object, undefined, w/e, but no parser at all will do the above? If EE is getting undefined or an object or something it's probably _not_ what the dev wanted anyway so a failure is probably fine.

OscarGodson:
With a simple "autocomplete" or whatever extension you could have it automatically finish the `>` after typing `<` or `"` after doing the first `"`. I can't wait for extensions! :tada:

johnmdonahue:
I went ahead and pushed the change since it's basically just a removal of the error message and pre tag wrapping we already discussed from the last iteration. Here's the code. It's pretty straight-forward.

``` javascript
if (!(typeof self.settings.parser == 'function' && typeof self.settings.parser('TEST') == 'string')) {
  self.settings.parser = function (str) {
    return str;
  }
}
```

Maybe we could just document that passing `parser: false` will disable the parser (effectively creating an html editor). I'm pretty sure this would work now. But to make things more clear if we end up emitting error events, we could do a simple check on the parser only if passed and `!false` and emit the error then.

johnmdonahue:
The possibilities for extensions are endless here!

EpicEditor: The Extensibly Embedable Editor &trade;

;)

OscarGodson:
Cool. See my pull request. I pretty much did the same thing :)

https://github.com/OscarGodson/EpicEditor/pull/118/files#L0R394

PR #116, opened 2012-05-22
Update - Utility Icons

sebnitu:
I've added the new icons and updated the JS to include the click even for both the preview and edit images instead of just the one since we're using different images now depending on the view.


OscarGodson:
Haven't pulled this, but just by looking does this work with the key shortcuts?

Whats the thinking for adding 2 click handlers instead of just one delegated `epiceditor-toggle-btn` click and check the app state to see what it should change to?

i.e. 

``` javascript
self.iframe.addEventListener('click', '#epiceditor-utilbar', function (e) {
  var targetClass = e.target.className;
  if (targetClass.indexOf('preview-btn') > -1) {
    self.preview();
  }
  else if (targetClass.indexOf('edit-btn') > -1){
    self.edit();
  }
});
```

Just thinking. Might not be the best, but this would cut the click handlers and safeguard against it somehow losing track of what is the right image to show.

sebnitu:
Well I'm hiding and showing the images with CSS based on the class of the wrapper. The reason I did 2 click handlers is because how it was currently working is only the first button was working, I think because it had `('epiceditor-toggle-btn')[0]` so it was only targeting the first element in the object? I tried just removing `[0]` to target the entire object but it threw an error so I added button specific classes to both images and targeted them that way with separate click handlers.

In short, I'm not actually swapping the images, just showing and hiding them.

OscarGodson:
Oh, so it should never lose it's state because the wrapper's class changes. Perfect. That makes sense.

For the `[0]` being removed and it erring, yes, jQuery takes care of that for you, but `('epiceditor-toggle-btn')` in vanilla JS is returning just a normal array and you can't set an event handler like click to an array, only an element, if that makes sense. You'd have to do a loop and add it like:

``` javascript
var el = self.iframe.getElementsByClassName('epiceditor-toggle-btn');
for (x = 0; x < el.length; x++) {
  el[x].addEventListener(function () {
    // codez
  });
}
```

However that's really bad for performance when you have lets say a 100 `<li>`s with some click handler. This is why you use delegated clicks like in my example above. The click handler is actually on the wrapper of the buttons and when you click the wrapper it checks if target of the click was a toggle button and if so does X. 

So here: https://github.com/sebnitu/EpicEditor/blob/71dbd652f66ef4641af574dc39dac6daeff63503/epiceditor/js/epiceditor.js#L507-L519

Could be changed to the preview comment code and have a single click handler so if we add a button we wont have to add another click handler which is also yet another click handler to keep track of. [More info](http://www.nczonline.net/blog/2009/06/30/event-delegation-in-javascript/)

Since the code is setting up a click handler to basically just do `.preview()` or `.edit()` could you change it to a delegated click handler? If you don't feel comfortable or feel it's complicated I can just pull it in and swap it, but my code example above should be about 90% of it (i didn't test it tho).

sebnitu:
No problem, I can update that. Makes total sense by the way. I don't work with raw JS very often so just need to keep this kind of stuff in mind :)

So if I want to make this change would I then create a new pull request with both commits and this one would just get closed, right?

OscarGodson:
No, just make your changes in the update/utilityIcons branch you made and push it. The pull request should automatically be updated. :)

OscarGodson:
Looks fab - U - lus

Gonna merge it in after I'm done with this stuff im working on. Really can't wait for these new icons!

PR #118, opened 2012-05-23
Better error handling

OscarGodson:
**THIS IS NOT FINISHED SO DON'T PULL IT IN!**

It's failing an unload test and not really all that tested by me. I just wanted to get a discussion going about if the following is even an issue or not. I feel it is, but don't have a good idea on how to fix it.

Let's say there's a parser error and you catch them like this:

``` javascript
var ee = new EpicEditor().load();
ee.on('error', function (err) {
  if (err.type == 'invalid_parser') {
    console.log('Parser is a broken because '+ err.message);
  }
});
```

You see the issue? Me either really aside from just knowing why this wont work because I wrote all the code for EE. The problem is the parser error already happened because `load()` was already called therefore the console log will never get logged like this even with a broken parser. You'd need to change it to:

``` javascript
var ee = new EpicEditor();
ee.on('error', function (err) {
  if (err.type == 'invalid_parser') {
    console.log('Parser is a broken because '+ err.message);
  }
});
ee.load();
```

It's not very intuitive tho. Not sure how to fix this or if it's event possible. Ideas? Is this even a real issue or should we explain that you should only put `load()` once _everything_ in EE is setup?

cc: @johnmdonahue


epiceditor/js/epiceditor.js
```diff
@@ -393,6 +388,15 @@
     , unloaded: false
     }
 
+    // Emit errors with parser before doing anything else
+    if (typeof self.settings.parser !== 'function') {
+      self.emit('error', {type: 'invalid_parser', message: 'Parser must be a function'});
+      self.settings.parser = function (text) { return text; };
```
OscarGodson, line 394: This has to be added or everything will crash...

johnmdonahue:
I think it's reasonable to expect users to add event handlers before `load()`. They would need to do this anyway if they wanted to hook into the `load` event. What if we abandon the idea of an error _event_ and just pass errors as the first param to callbacks. Following the Node convention, maybe we could do something like this:

``` javascript
ee.load(function (err) {
  if (err && err.type === 'parser_err') {
    handleError();
    return;
  }
  console.log('EpicEditor Load Ok... ')
})
```

Also by default `load()` goes into editor mode, so this would work there as well if we pass the error to each listener. Something like this:

``` javascript
ee.on('load', function(err) { })
ee.on('preview', function(err) { })
ee.on('fullscreen', function(err) { })
```

Thoughts?

OscarGodson:
The code you have above will still fail because the load event/callback is fired _after_ it's all loaded and the parser will have already failed. The callback and the emitted `load` event are fired the same time. For the above to work the load callback would need to be called "beforeOnLoad" which doesn't make sense in a callback since callbacks usually mean "all done". I _do_ expect the above to work tho.

The only issues with err as the first param is most people will ignore errors and there's not really any errors built in. I feel like most people are going to come from a jQuery style of thinking where the first param is the data like (why currently `file` is the first param):

``` javascript
$.ajax('/foo', function (data, xhr) {
  if (xhr.status == 'error') {
    throw new Error("Dammit");
    return;
  }
  $('#content').append(data.content);
});
```

What types of errors do you foresee? Here's what I can think of: `invalid_parser`, `not_loaded`, `failed_to_load`, and `storage_error`. These all need to be fully implemented.

What do you think? Should we appeal to more cautious, handle failure Node type or the "i expect this to fail silently" jQuery type?

OscarGodson:
@johnmdonahue I think I'm gonna punt this for a later release. I like just falling back silently just like it is now quite a bit and I'd like to get a bunch of error checks rather than just a couple before releasing.

johnmdonahue:
Definitely in agreement on leaving this open for now. It would be good to get feedback on what kind of errors people are actually running into as well and how they would expect to be able to handle them.

OscarGodson:
Closing. This can't ever be merged now. Should revisit tho.

PR #119, opened 2012-05-23
Bugfix/editor state

sebnitu:
Fixed a bug where the previous state of the editor was not being displayed after leaving full screen view.


OscarGodson:
Wow simple fix. Was thinking I had to manually update the classes. I'll check this when I can! Thanks a lot! @dadambickford looks like you don't have to do this.

OscarGodson:
Dropped it in! Worked like a charm!

PR #123, opened 2012-05-25
PUBLIC BETA!

OscarGodson:
I did a quick run through and checked IE9 and everything seems to be functional. Don't pull this in yet. I want to run the test suite in IE9 first. But getting this stubbed out.

Does anyone see anything that needs to be fixed? Are the docs good? Any major bugs? @johnmdonahue do you wanna get any last minute stuff in?

QA'd by @dadambickford


OscarGodson:
One thing we need to not forget is to update the image URL to work in the README to be `http://epiceditor.com/images/epiceditor-logo.png` so when we push this to master / gh-pages it'll work.

OscarGodson:
## Chrome, Firefox, and IE9 are all passing the test suite now

tests

:shipit:

OscarGodson:
Looks like we're already to go, but there's 2 bugs remaining that are kinda shitty.

The first is that _sometimes_, and I can't repro reliably, the shortcut keys block the input in the editor and trigger the shortcuts even without `alt` being pressed. I.e., `f` once got "stuck" and whenever I pressed it fullscreen opened. `O` also did it once for me, but since the editor was already opened I just couldn't type `O`.

The other is with semi-big docs like: https://gist.github.com/1d478af940c72e0f6a9e sometimes there's dropped keys. In other words, if I type: `123456789` on the end of the `better interval cleanup` line for example, half the time the `9` doesn't show up until I press another key. 

I want to fix the 2nd issue before releasing. Other than that I think we're good to go.

OscarGodson:
OK, fixed the 2nd issue. Couldn't sleep so I just did it :wink:

OscarGodson:
Also, don't forget, after we put this in master we should make a develop branch and start working in develop

OscarGodson:
This has been merged in.

PR #142, opened 2012-07-05
Set images path

istana:
Hi, I modified editor so you can set path to images. I find it useful, when I want have css, js and images directories first and then epiceditor folder (if it is neccessary)
Lint and build passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ var opts = {
     preview:'/themes/preview/preview-dark.css',
     editor:'/themes/editor/epic-dark.css'
   },
+  images: '/images/',
   focusOnLoad: false,
   shortcut: {
     modifier: 18,
@@ -142,6 +143,11 @@ var editor = new EpicEditor(opts);
     <td><code>theme.preview</code></td>
     <td>The theme for the previewer.</td>
     <td><code>themes/preview/github.css</code></td>
+  </tr>
+  <tr>
+    <td><code>images</code></td>
+    <td>Path to images.</td>
+    <td><code>/images/</code></td>
   </tr>
   <tr>
     <td><code>shortcut.modifier</code></td>
--- a/src/editor.js
+++ b/src/editor.js
@@ -289,6 +289,7 @@
           , preview: '/themes/preview/github.css'
           , editor: '/themes/editor/epic-dark.css'
           }
+        , images: '/images/'
         , focusOnLoad: false
         , shortcut: { modifier: 18 // alt keycode
           , fullscreen: 70 // f keycode
@@ -300,6 +301,10 @@
       , defaultStorage;
 
     self.settings = _mergeObjs(true, defaults, opts);
+    
+    if (self.settings.images[self.settings.images.length-1] != '/') {
+      self.settings.images += '/'
+    }
 
     if (!(typeof self.settings.parser == 'function' && typeof self.settings.parser('TEST') == 'string')) {
       self.settings.parser = function (str) {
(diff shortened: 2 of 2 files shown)


OscarGodson:
I think that's a great idea and probably would be good to do the same with CSS too.

Could we make a new `path` group of settings? I think it'd make more sense and we could keep building it up (maybe at some point we'll want to include external JS or a plugin directory?):

``` js
var editor = new EpicEditor({
  path: {
     base: 'epiceditor'
 ,   images: '/images/'
 ,   css: '/css/'
 }
});
```

That way you could setup your base to be a blank string and you'd be able to include your images and css in your normal site directories.

thoughts? And thanks for running the lints and tests! :thumbsup:

cc: @johnmdonahue

johnmdonahue:
Awesome idea! A few comments/questions:
1. Would it make sense to possibly rename `path` to `assets` in this context? 
2. As @OscarGodson mentioned, I can imagine a scenario where a user may want to overwrite just the `/images/` path but keep the original JS in some default location (`/epiceditor/js`). If we just remove `base` it might save some confusion. Thoughts?

``` javascript
var editor = new EpicEditor({
  assets:
    { js: 'epiceditor/js/'
    , images: 'epiceditor/images/'
    , css: 'epiceditor/css/'  // Is this the path to themes?
    }
})
```

Example:

``` javascript
var editor = new EpicEditor({
  assets: { images: 'custom/images/' }
})
```

OscarGodson:
@johnmdonahue base is actually pretty useful. In our tests we use it for example and when people send me JSBin's or I share JSBin's with them I can just do basePath. In this case I'd have to update everyone to be like:

``` js
var editor = new EpicEditor({
  assets:
    { js: 'https://raw.github.com/OscarGodson/EpicEditor/develop/epiceditor/epiceditor/js/'
    , images: 'https://raw.github.com/OscarGodson/EpicEditor/develop/epiceditor/epiceditor/images/'
    , css: 'https://raw.github.com/OscarGodson/EpicEditor/develop/epiceditor/epiceditor/css/'  // Is this the path to themes?
    }
})
```

Might not be too big of a deal since it's mostly for testing and development, but it's useful if you are hotlinking. I guess you could store the "base" in a var too and do like `js: base + '/js/'` tho too. Don't feel too strongly either way, just a thought.

And yes, the css directory is the themes.

I also like `assets` better. Makes it sound more like you can list arrays of data like `plugins: ['plugin.js', 'blah.js']`

johnmdonahue:
Ahh gotcha. Either way works for me as well. My only concern was that someone might try to overwrite the `images` path _alone_, for example, and be scratching their head about why `epiceditor` was still being prepended.

Also, did I see mention that today is your birthday?!? If so, hope you're having a great one!

:birthday:

OscarGodson:
That's a very valid  valid concern. @istana what are your thoughts on this? What would make sense to you as someone who hasn't been working on this all the time and doesn't know the entire API by heart? :P

Thanks for the birthday wishes too btw!

istana:
I find `basePath` useful, because I usually have 'assets' on one domain. (btw. I like renaming `assets` and `/themes/` part to `/css/`) If somebody will have images on another location? Then basePath should be empty and and css/images/js should have full absolute/relative path. It's only three items, so it's not daunting task.

OscarGodson:
OK, I put this on the 1.0 milestone. @istana Do you want to go ahead and make an `assets` group of options and put `images` inside of that? We can also remove `basePath` I think because as @johnmdonahue pointed out it might be confusing. 

We'll need to move the `themes: {...}` to `css` too and have it work without `basePath`. I'm not sure how to handle the CSS stuff tho. Does it become:

``` js
assets: {
  css: {
    theme: {
      base: '/themes/base/epiceditor.css'      
      , preview: '/themes/preview/github.css'
      , editor: '/themes/editor/epic-dark.css'
    }
  }
}
```

???

istana:
yes, I will do it
imho you can omit theme in `css` if there are no plans to have styling other than theme

johnmdonahue:
@istana: +1 on simplifying.

Just thinking out loud here but what do you guys think about setting the icons as css backgrounds rather than `` tags? This way users can just overwrite the style to modify buttons etc and not have to worry about setting paths to images. This could simplify the whole setup and remove the need for assets altogether. We can always extend `js` stuff later if we want to expose some sort of plugin api - i.e. `new EpicEditor({ plugins: [] })`.

``` javascript
var editor = new EpicEditor({
  theme:
    { base: '/themes/base/epiceditor.css'
    , preview: '/themes/preview/github.css'
    , editor: '/themes/editor/epic-dark.css'
    }
})
```

OscarGodson:
@johnmdonahue +1 on  images as background images and leave theme and basePath the same, plus this wont break anyone's editors on the 0.1.2 upgrade, but change the image paths in the actual theme. I mean, it is part of the base theme anyway. 

I also like the idea of plugins handling their own assets in their own way and you just include them like you have above.

OscarGodson:
@istana Any updates?

istana:
@OscarGodson Sorry for not responding so long....I was somewhat burnt out
I like idea, that images will be linked in CSS file. "Normal" images as background images are necessary evil.

I modify it in few days

OscarGodson:
No rush! and thank you!

OscarGodson:
It's been a long time and no update so closing. Feel free to reopen or submit the changes you mentioned :)

PR #151, opened 2012-07-13
Add load event

fntz:
Hi.

I changed `load` method, because it is not possible to work with the editor before the DOM was not loaded. Now, without any additional libraries, you can use this method. 

Before
(with jQuery)

```
$(document).load( funciton() {
});
```

Now the wrapper is not required:

```
```


OscarGodson:
Thats a big commit. Give me some time to go through it and test it.

A few questions:
1. In what cases would you need this? Seems like this would only be a problem if you put script tags in your `<head>` which, for DOM specific stuff, is a bad idea anyway.
2. Do all tests and lints pass?
3. Is there any specific reason why you're using a double underscore and not a single one to denote a private method/variable?

Thanks for the pull request! Much appreciated.

fntz:
Hi.

1) I include all scripts in HEAD tag. But when i include `EpicEditor` then i got error `document.body.webkitRequestFullScreen is undefined or null`. Then I created new branch to people who are included scrpipts in the HEAD tag, did not got an error.  I agree that it is necessary only in that case do not use third-party libraries.

2) All tests were completed successfully. I added new parameter to `open` method, because when I did the tests, it do not pass just because the editor is not loaded.

3) I use a double underscore for private function.

Also, I found a bug, although it is most likely a bug browsers. 
When refresh page with the tests, it turns `localStorage` not cleared. But when you manually delete all of the localStorage that all tests pass.

Thanks.

fntz:
Remove this реквест because inclusion in a code of the editor of dependence on loading, will create more problems. It is more correct to use foreign libraries or yourself to control the loading document.

Thanks.

OscarGodson:
I closed the pull request as per your request, but my first question was more, why are you putting scripts in the head?

The open method should just return early if the editor is not loaded. There should be some stuff in the code now that handles that. You shouldn't have to add a param to open, it should handle it it's self.

We use a single underscore to denote privates, so stick with that convention to keep the code all the same :)

Thanks!

Issue #144, opened 2012-07-09
Alt-O not working (OSX)


pawelphilipczyk:
Hi, 
Alt-O (Opt-O) combination is not working for ó and Ó which is Polish character under this shortcut. 
I guess in otherl languages this might be the issue as well.

OscarGodson:
If you're doing alt+o then yes, it'll catch that key and prevent the input. This is tough because I'd love to have shortcuts and I know a lot of people use them, but no shortcut is going to work across the board (i think).

If you're adding this to your site and it gets mostly Polish visitors you could disable the shortcuts via the options.

Do you have any suggestions on shortcut combos or other way's to fix this?

pawelphilipczyk:
If I can disable the shortcuts then there is nothing more I'd suggest.

This is fantastic editor and thank you for making it!

pawelphilipczyk:
After a moment of using the shortcuts I do have a suggestion that may be not only fixing this problem but also be a usability improvement.
The Alt-P for preview could switch edit/preview mode not just switch to preview mode. So there would be just one shortcut combo for toggling this.

OscarGodson:
Great idea! It also makes sense since we have a single preview/edit toggle button. I'm going to put this on the 1.0 milestone and try to get it into the `0.1.2` release.

OscarGodson:
@pawelphilipczyk can you pull develop? @jeffhill actually sent a pull request for this and fixed it! I just pulled it in and tested it and it appears to be working good for me :)

pawelphilipczyk:
I've pulled it and tested it. The Alt-P works fine and Alt-O no longer blocks my ó character. I have other problem but this might be not related to EpicEditor at all. I'm using it with Drupal and [EpicEditor](http://drupal.org/sandbox/erikwebb/1674176) module. After hitting Alt-P the textarea loses focus and to switch back to edit mode I have to click it before I can hit Alt-P again. I'll check that with module maintainer over there but maybe you have any thoughts on that? Here's module [issue](http://drupal.org/node/1690726) link on drupal.org.

CashWilliams:
I'm cross posting this from the Drupal issue linked above.

What I'm seeing is that Alt-P enters Preview mode, but pressing P again (without lifting Alt) does not return to edit. However, if I lift all keys, and then press Alt-P, I will return to edit mode.

This may be a limitation of the OS?  (OSX 10.7 for me)

pawelphilipczyk:
[...]

PR #154, opened 2012-07-17
Ticket #144 - Alt+P now toggles between preview and edit.

jeffhill:
I think this takes care of ticket 144, so that Alt+P toggles between preview mode and edit mode.


OscarGodson:
Rock on man! Thanks a lot. Why not remove the alt+o stuff while you're in there too. So, remove the info from the README, the option from the options object in load() and all the alt+o code below the alt+p code? If you can't get to it I'll take what you have but that'd be a huge help :)

Thanks again!

jeffhill:
This commit takes care of removing the Alt+O shortcut and removing the shortcut.edit setting from the options object.  I've also updated the readme to reflect these changes.

Thanks for the opportunity to contribute!

OscarGodson:
Pulled it in @jeffhill thanks a lot! It all looked great. :thumbsup:

Issue #164, opened 2012-07-24
In Firefox you can't type HTML tags


OscarGodson:
I was trying to test another thing and I typed some HTML like `<h3>FOO</h3>` and previewing and going back wiped the HTML out because of how we shimmed `innerText` for Firefox.

OscarGodson:
Fixed from hotfix 0.1.1.1: 78a80c884f7f5b2174864cc67f7f5990dea726b9

Issue #159, opened 2012-07-20
TAB key to indent/insert tab character


hongymagic:
Hi,

I have hacked up initial version of `TAB` insertion in [my own fork](https://github.com/hongymagic/EpicEditor/commit/efdee33ec96e66968415658becfae3e2a890b5f0) (note: it's initial and does not support `shift-tab` combination yet).

I just wanted to start up a discussion around supporting `TAB` keys. The reason why I ask is because it has come to my attention that not everybody agrees with `TAB` key being used to indent; especially in the context of a webpage. _So what do people think about adding support for `TAB`?_

Ideally, in the context of an editor, I would use `TAB` key to indent selected lines or for (de)indentation. We can make use of [tab-size](https://developer.mozilla.org/en/CSS/-moz-tab-size) CSS property to adjust tab-width.

OscarGodson:
I for one love this idea. :+1:

I wanted to get this in there eventually. I'm almost certain people will want to disable this tho for some reason or another. Why not just throw a `smartTab` option in there? `smartTab` could default to `true` and "smart" tab because it makes tabs work like a real text editor.

It'd also be super awesome if you could get it so it remembers the indentation of the last line. So, if the `|` is the cursor:

```
text...
    a tabbed line and I press enter the cursor ends up...
    |
```

Anyway, when you think you're ready send a pull! This is looking great. Curious, why are there a bunch of empty lines added/deleted?

hongymagic:
As @OscarGodson said:

> Why not just throw a smartTab option in there? smartTab could default to true and "smart" tab because it makes tabs work like a real text editor.

That is one way to handle differences I guess! I will gladly add another option `smartTab`.

> It'd also be super awesome if you could get it so it remembers the indentation of the last line.

I agree this would good too. Will definitely take this into account.

> Curious, why are there a bunch of empty lines added/deleted?

I noticed that too, I am assuming my vim settings didn't match EOL of the file. Don't worry, I will create a new branch and make sure current code/file formatting rules are followed.

Thanks!

OscarGodson:
No worries! Was just curious. Do you think it should be plural or singular? `smartTab` or `smartTabs`? I'm leaning towards plural now that I say it out loud.

hongymagic:
> Do you think it should be plural or singular? smartTab or smartTabs?

`smartTabs` should do the trick =)

> Curious, why are there a bunch of empty lines added/deleted?

[...]

Issue #41, opened 2012-03-06
Optional buttons for common styling


OscarGodson:
Sort of like a WYSIWYG, have buttons for bold, italics, links, etc. see GitHubs wiki editor for an example. This should be off by default tho.

librarian:
I need this feature. Maybe you could suggest how to implement this feature by myself?

OscarGodson:
@librarian This will be tough and why it's a v2 thing. The tough part is moving the cursor around. It's a contenteditable so it'll be like implementing any other WYSIWYG buttons. I suggest googling it because there's a lot you'd have to do. You could have a look at how markItUp does it: http://markitup.jaysalvat.com/examples/

If you do it, let me know. I'd be interested in your implementation.

librarian:
Thanks, I will try at my free time

mahemoff:
This would be great to see. It looks like Rangy is the way to go (http://code.google.com/p/rangy/wiki/SelectionSaveRestoreModule). Relevant: http://stackoverflow.com/questions/1181700/set-cursor-position-on-contenteditable-div

OscarGodson:
@mahemoff @hongymagic was really close then left on vacation. He actually had the selection AND a prototype with buttons. The problem was the scope of the pull request was too big. If he makes a new pull request with just the text selection stuff I'll pull it in. If I don't hear from him by `0.3.0` i'll start looking into getting some text selection API's in myself.

Original Pull: https://github.com/OscarGodson/EpicEditor/pull/166
Demo of buttons: http://epiceditor-demos.herokuapp.com/

corydeppen:
Would you be able to provide an update on the status of this request?

OscarGodson:
@corydeppen Unfortunately, I havent had time to keep up with this project lately. This was supposed to be for 2.0 but there's still a lot to do for 1.0 :( https://github.com/OscarGodson/EpicEditor/issues?milestone=3&state=open

PR #166, opened 2012-07-29
Enter key now inserts `\n` instead of leaving it for browsers to handle them

### LINKED ISSUE #162: Injecting content/markup at cursor location
Would be an awesome feature to have. I know I would put it to good use :)

hongymagic:
This change is in preparation for issue #159. I think this code is ready for review. Coincidentally, this fixes: #162 and #164.

You can see the initial implementation of #159 (TAB/indentation support) on [my fork branch](https://github.com/hongymagic/EpicEditor/tree/feature/indentation). It was branched off this pull request.

---

Things to look out for:
1. Public API (there are 3 new public API calls)
2. Event handling
3. Specs/Tests… Please review the tests. Testing `insertText` was a little tricky without way to get text from EE.

---

There is an issue with Firefox at the moment, whilst the `insertText` works when run in browser, the spec/tests I wrote fail in Firefox. It doesn't seem to like the `getSelection` call. I am currently trying to look at why and how to fix it. If you could help me polish the tests, that would be super helpful – as this is a big pull request.

I have tested this code in Chrome, Firefox, Safari, and IE9.


OscarGodson:
Here's what I noticed so far:
1. Text no longer wraps in Chrome or Firefox at least. I didn't check the others yet. screenshot
2. cmd+z no longer undos.
3. `findOffset` is a private function, can you add an underscore?
4. Why use `getText` and not `exportFile`? `exportFile`, by default, exports the exact same thing. Unless you have something else planned it appears you could remove that and use `exportFile`.
5. Any particular reason you have the private methods in the load method and not out with the others? If not, could we hook it as a sibling to all the other methods? It'll make it easier when we split the library into multiple files also it wont have to reinit those functions whenever the editor is `loaded` and `unloaded` which some people do a lot in their apps.

I'll take a look at the specs and stuff next :) Looking great, and `insertText` is sweet. I'm sure @sebnitu will be happy. This will also make it super easy to do #41.

OscarGodson:
Oh, and what is the event handling stuff you wanted me to see?

hongymagic:
> Text no longer wraps in Chrome or Firefox at least. I didn't check the others yet.

I added `word-wrap: normal` to themes/editor*.css files. I will remove them. And I also need to enable `white-space: pre` in the load as well as you mentioned in the gist, at the moment all styling changes are inside these css files.

> cmd+z no longer undos.

… It looks like Chrome only supports undo in `innerText/innerHTML` at a fairly granular level, Firefox... [see this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=290261). The gut feeling tells me that Firefox doesn't like calls like `textContent = ''` or `innerText = ''` as it removes the DOM tree. I will take a closer look at this issue in Firefox.

Since `insertText` and newline uses `_setText` which in turn need to set `textContent/innerText`… if Firefox doesn't support undo on those, the only solution I can think of at the moment is rolling our own UndoManager. Probably an overkill at this stage imo.

> Why use getText and not exportFile? exportFile, by default, exports the exact same thing. Unless you have something else planned it appears you could remove that and use exportFile.

`exportFile` was giving me some grief, but I will try it again.

OscarGodson:
I actually only tested all this stuff in Chrome, so the undo thing im talking about is in Chrome. I guess its not a huge deal and we should see if it's ever reported by someone else. Well be able to close a bunch of tickets with just this pull request so I think it'll be fine.

What was happening with exportFile?

hongymagic:
> Oh, and what is the event handling stuff you wanted me to see?

I just wanted to double check if you wanted new events for `insertText`. But I guess `autoSave` will trigger `save` anyway.

> What was happening with exportFile?

I was getting empty strings, but it turns out I wasn't saving before exporting. I've updated the api/specs already.

> could we hook it as a sibling to all the other methods?

This makes perfect sense. No reason why every instance should have its own. I've refactored it out with methods such as `_getText` and etc. 

Just doing some final specs + documentation changes!

OscarGodson:
Undo seems to be working better! One blocker tho on the pull is everything moves to one line when you hit enter. To repro, clean out your localStorage (`localStorage.clear()`) then hwhen you get the default text for the docs: 

``` md
#EpicEditor
This is some default content. Go ahead, _change me_.
```

After the `.` type gibberish words like `fljsdfj fjlskfj lsjf ldsk` until the text wraps a few times. Hit enter. Now you should see this:

``` md
#EpicEditorThis is some default content. Go ahead, _change me_. hfsdkj fhksdf hksdjfh skdjfh kdsjhf skdjfhdks jfds kjfh dskjfhdks fhsdk fhskjfhskjfkjdks hf kdjsf kjsdf dksj ksjf ksjhf kjdsfhkjsdfksdjf kjdshf kjsdfhkjsdfhkjsdfhkjdsfhdksdskfds
```

But otherwise stuff seems to be looking better!

hongymagic:
Blocker mentioned above should be fixed with the latest push. I have also added two items into README.md as per EE's contribution guide.

I am also very tempted to change `_setText` to use something like:

``` javascript
_setText = function (el, content) {
     var textNode = el.firstChild;
     textNode.nodeValue = content;
    return true;
}
```

Purely because undo command works better with it. The only problem is ensuring that the `firstChild` is indeed a TextNode, and if it wasn't – looping through child elements to find first TextNode may be too slow or even unreliable.

Let me know what you think.

hongymagic:
Also very happy to fix what ever bugs come up around text inserting, newlines and etc in the future.

OscarGodson:
As long as you can reliably get the text node element I think it's fine. I did a super simple [JSPerf test](http://jsperf.com/innertext-vs-firstchild-nodevalue) to compare the speed of each and firstChild nodeValue is a tad slower in Chrome, but nodeValue is twice as fast as innerText in IE9 (and probably IE8). Go figure...

So, might even be better in the long run because Chrome can take a slight performance hit, the IEs could use every drop of performance enhancements they can get.

And, glad to hear about the future support :) I make sure to tag you in tickets if/when they come up after this pull is finished. It's getting closer! I'll take a look tonight.

OscarGodson:
Hows this coming?

hongymagic:
This is ready for last round of review! I've changed the setText in the last commit and that should complete the newline work for now.

OscarGodson:
Noticed some weirdness. In Chrome, the text seems to wrap whenever, but it should wrap per word, not character. 

Your branch:



On EpicEditor.com (master):



Notice "typing" is wrapping at the "y" in the first image, but should wrap like the 2nd image where the whole word wraps.

In Firefox it doesn't wrap at all:



I'm going through your code and documentation now :) I didn't test the above tho in IE, but make sure IE9 also behaves correctly in the above cases.

README.md
```diff
@@ -230,6 +231,20 @@ openFileBtn.onclick = function () {
 }
 ```
 
+### getSelection()
```
OscarGodson, line 234: This still needs a code example. All API methods should have an example in the docs. Even if it's something as simple as:

``` javascript
var selection = editor.getSelection();
console.log(selection.baseNode); // the text that is highlighted
```

README.md
```diff
@@ -230,6 +231,20 @@ openFileBtn.onclick = function () {
...
+
+Get current selection from the editor. Object returned is same type as `window.getSelection()`.
+
+### insertText(_text_)
+
+Inserts given string of text into the editor. Any selected text will be replaced with the new text. Otherwise text is inserted at the current caret position.
+
+Use `\t` for tab, and `\n` for newline.
```
OscarGodson: Can you make this "Use \t for tab..." a "Note:" like in enterFullscreen or move it to the paragraph above? Looks odd with just that little bit of text is all.

epiceditor/js/epiceditor.js
```diff
@@ -1002,6 +1269,25 @@
   }
 
   /**
+   * Get current selection
+   * @returns {Object|Null}
+   */
+  EpicEditor.prototype.getSelection = function () {
```
OscarGodson, line 1230: Will this work across browsers? In other words, how with this code can I get the text of the selection? Above I see a feature check, but are the objects totally different in IE compared to Chrome and Firefox? If so you might want to put those examples in the getSelection docs.

spec/spec.js
```diff
@@ -555,9 +557,37 @@ describe('EpicEditor.rename', function () {
...
+    editor = new EpicEditor({ basePath: '/epiceditor/', container: testEl }).load();
+    emptyFile = 'emptyFile' + _randomNum();
+    editor.importFile(emptyFile, '');
+  });
+
+  after(function () {
+    editor.unload();
+  });
+
+  it('check to see if insertText inserts correctly', function () {
```
OscarGodson: If it's possible, is there a way to check if the text is actually inserted _after_ the cursor position as well? One of the biggest use cases for this is like adding things like a MD image tag or something where you're typing so:

```
Starting text. |

```

If the `|` is my cursor and I do `insertText('End')` it should put `End` at the end of the line. Can we check the indexOf the inserted text and make sure it's in the correct place?

epiceditor/js/epiceditor.js
```diff
@@ -143,34 +211,196 @@
...
+    if (selection.rangeCount) {
+      return _getSelectionStart(iframeDocument) + (selection.getRangeAt(0) + '').length;
+    }
+
+    return 0;
+  }
+
+  // Insert new line and respect smart indentation if the option is present
+  function _insertNewLine(iframeDocument, smartIndent) {
+    var body = iframeDocument.body
```
OscarGodson, line 194: newline is one word so it should be `_insertNewline`.

OscarGodson:
So, all my code comments are pretty minor which is great news! I'll probably let @johnmdonahue weigh in before I merge it since this is such a huge change after you finish all the bug fixes.

One actual code thing, we need at least one test on every public API method, so `getSelection` needs at least 1 test, preferably more.

Also, one more blocker, this is happening in Firefox (everything is an h1?):

hongymagic:
Simple screens of EE running on:
- Chrome 20
- Firefox 14
- IE 9

http://db.tt/MiCT8zdl

hongymagic:
@OscarGodson RE: Firefox rendering everything as `h1`

This may be related previous bug where it saved to localStorage without newlines. Could you run `localStorage.clear()` and try again? Cheers!

hongymagic:
> @OscarGodson wrote:
> So, all my code comments are pretty minor which is great news! I'll probably let @johnmdonahue weigh in before I merge it since this is such a huge change after you finish all the bug fixes.

Yes this would be good. I fear that this pull-request is become too _epic_.

Only thing I need to do is extending tests for `insertText`. I'll think about how I can test this with different caret positions. The problem I foresee is the need to set "caret position" programmatically. May have to write something similar to the private method `setSelectionRange` in order to set caret position, and even then, some browsers don't allow this behaviour!

OscarGodson:
Took a quick look, looking better. Almost there! Getting these errors tho in Firefox at least. Didn't see it in Chrome tho.

OscarGodson:
Looks like it happens whenever you press the key command (alt+p) to go to the previewer.

johnmdonahue:
I definitely dig the work towards a `smartTabs` option and the `smartIndent` action! :thumbsup:

Let me know if I'm leaving anything out, but to summarize the scope of this PR:
- Add a `smartIndent` option
- Add 2 new public methods: `insertText` and `getSelection`

### Questions
1. RE: `insertText` and `getSelection`: I know @sebnitu asked for insertText so maybe he could weigh in here as well, but what is the intended use case for the new public methods? How do you imagine these being used in the wild? I'm legit just curious :) Is the idea here to eventually hijack native copy/paste?
2. Is there any reason the `getSelection` method shouldn't just return the string itself? Again I think a use case will help me understand this better, but I can imagine a _selection object_ confusing users.
3. Would it be clearer in the documentation to use `.toString()` directly rather that `.call` - e.g. `editor.getSelection().toString()`

### Mystery bugs

I am seeing some intermittent strange behavior coupled with the following error in Chrome OSX: `Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1` pointing to line 227: `range.setStart(textNode, ss);`. I'm not sure how to reproduce it consistently, but I am seeing the following issues regularly:

**Please let me know if anyone else can replicate this**
1. Unable to use the arrow key to move _down_ - cursor jumps back up - unless I arrow right.
2. Up/down arrow movements seem to cause this error often - for example repeatedly pressing the arrow key down on the last line of text or pressing an arrow key after pasting text.
3. Selecting text to delete, copy/past/undo/redo manipulations, as well as pasted text grabbed from the docs will do any or all of the following:
   - create multiple duplicates of the selection or entire document
   - add new lines
   - result in the above error

I was using the `feature/newline` fork to write this and got up to 77 errors and 4 _different_ copies of the document in the editor before I needed to switch to another editor. This is with cleared localStorage and refreshes. I didn't notice my doc had jumped from from 21 to 94 lines until I started scrolling.

Thanks for all the hard work so far @hongymagic! I hope we can figure out what's causing the issues and get this all sorted. Maybe it's just me :) I haven't had a chance to go through the code much yet or test in anything other than FF/Chrome OSX, but hopefully this is enough of a start for now.

Let me know if you have any questions for me. Thanks!

OscarGodson:
@johnmdonahue 

For #1 insertText is a huge win. You can start to finally add WYSIWYG like buttons finally. Or, you can click an image in a modal and the MD image code with the image src is automatically inserted into the editor. There's a bunch of use cases, but I think those would be the clearest use cases.

For number 2 I also have had the same question. It's in a comment somewhere... but, I was thinking the same thing. I also made the point that it needs to "just work" across browsers like `enterFullscreen` for example. `getSelection` shouldn't just return the raw selection object since I don't believe that works consistently across browsers and the end developer will need to understand the browser selection API's which are über confusing.

The selection API method should, in my understanding, make it so a developer can easily do things with a selection. So, for example, be able to get the selection, and wrap it in `<strong>` or if you selected `http://google.com` and clicked a button it'd take that URL and make it into a MD link. @hongymagic How could I do those with this API method?

For #3, I also agree. I like that!

I'll take a look @johnmdonahue at those errors you're seeing tonight.

johnmdonahue:
Ahh! Gotcha! Yeah, I figured there was some obvious use case there - I wasn't thinking along the lines of WYSIWYG but that makes perfect sense. With that in mind, maybe the docs could have a blurb (e.g. "super-useful for adding WYSIWYG-like tools") or just show an example of wrapping a selection to italicize etc, so that it's purpose is more apparent.

OscarGodson:
Yeah, that'd be a great idea. I've gotten numerous requests for this. And I agree with them, having to teach a client how to use MD isn't realistic, but at the same time I wanted the core editor to be barebones. If the API tools are easy enough I'd love to write out a basic button toolbar for MD syntaxes as a sample that others can work off of. Or, an extension, but we still haven't really talked about those yet.

hongymagic:
A very simple example of turning selected text into bold text:

_Select a word or sentence_, then:

``` javascript
var s = editor.getSelection();
var text = '**' + s.toString() + '**';
editor.insertText(text);
```

[Before image](https://www.dropbox.com/s/91yajpa45kib1dl/001.png)
[After image](https://www.dropbox.com/s/tk05ca2hh9oytbh/002.png)

OscarGodson:
WOW if that's seriously it that's pretty fucking amazing and I'm super excited! Put that exact sample in the docs. That's rad!

P.S. if you saw the typo, it was autocorrect. I hate you autocorrect.

hongymagic:
May I suggest running a feature branch based off this and implement/test things like indentation, copy and pasting? Without these features, it will be hard to take this base as a replacement just yet.

What do you guys think?

EDIT: I am also travelling from September for two-whole months \o/ so I won't have much time to build upon this for a while. If we could stash this on a branch and get it concrete and implement TABS/copy and paste – then I think we could safely merge it into the develop branch :)

OscarGodson:
As it is now with just the fake new lines and the two API methods id love to merge it. I don't think we need to add those other features as requirements for this pull request. This alone will be incredibly useful and close out at least 3 tickets I know of off the top of my head. Also, I've been getting more and more people requesting the ability to add WYSIWYG buttons and they need this to do that. I can finish any docs changes or maybe even finishing the tests if you can get this in to a non-error, IE supported state :) additional tests and the doc changes we mentioned would be a plus, but not required.

But yeah, those bug fixes would be awesome!

hongymagic:
I have added [additional documentation](https://github.com/hongymagic/EpicEditor/commit/7675f3432d890fa5b379a6d6f2dd3b167353fb22) and also migrated old tests to mocha plus [new tests for insertText](https://github.com/hongymagic/EpicEditor/commit/203369b7abbd6ba475be1c1cd9ef32d4e086d90c). See my last push (6 days ago).

This works fine on Chrome/Firefox and IE9. What level of IE support were you after?

johnmdonahue:
@OscarGodson @hongymagic Are either of you guys seeing the issues I mentioned? Or is it just me? I can definitely take another look if so.

OscarGodson:
Oh cool! I'll check this out then. I'm on mobile github issues so I can't see your commits in this feed.

OscarGodson:
And yes, IE9 is fine for now :)

OscarGodson:
Take another look on a couple browsers if you could. Can't test at the
moment  :)

hongymagic:
@johnmdonahue sorry about not replying about those mystery bugs. I'm frequently getting lost in this comments chain :P

Mystery bugs

> INDEX_SIZE_ERR: DOM Exception 1 pointing to line 227: range.setStart(textNode, ss);

I believe I've ironed this out in [this commit](https://github.com/hongymagic/EpicEditor/commit/428aec5385541cad792b2404c9d5d4c5c70ae58b). This error occured frequently on all browsers esp on empty documents.

> Unable to use the arrow key to move down - cursor jumps back up - unless I arrow right.

Unable to reproduce now, probably because it was related to above error/fix.

> Up/down arrow movements seem to cause this error often - for example repeatedly pressing the arrow key down on the last line of text or pressing an arrow key after pasting text.

Pasting (same as undo) is going to cause a lot of issues with this particular style of text manipulation unless we implement our own paste function :( This is because browsers' tend to paste the HTML inside the body as well – when we expect our body to be HTML free. Paste should be relatively easy to implement though :)

> Selecting text to delete, copy/past/undo/redo manipulations, as well as pasted text grabbed from the docs will do any or all of the following:
> - create multiple duplicates of the selection or entire document
> - add new lines
> - result in the above error

Same point as one above. This is why I wantd to experiment with our own cut/copy/paste, indentation, and undo manager on a separate feature branch. But as @OscarGodson mentioned, those can come later :)

OscarGodson:
@hongymagic So here are some kinda bad bugs I found:

1 - If I paste _raw_ text it's missing all the formatting. This is pretty shitty since a lot of people paste text into the editor to try it out. Also Seems to be with not being able to copy the new lines? This doesn't happen in master:



2 - I can't paste elsewhere like in this post after copying either... :(

3 -. Make a bunch of lines like:

```
fdsfdsfs
f
sdf
sdf
ds
f
ds
ds
```

press and hold shift, press the up arrow. It doesn't highlight all the lines like it does natively. This makes it impossible to select chunks of text on multiple lines. The odd thing is that it works fine pressing down. Hm.

4 - When I press enter at the end of the document it doesn't scroll. The cursor just keeps going down the page and the scrollbar grows, but it needs to scroll too.

Undo/redo seems fine to me, @johnmdonahue but copy and pasting from the editor to elsewhere is really janky and copying and pasting from the editor is a pretty standard thing. We'll wanna get that part right along with the above bug fixes first before pulling.

OH! Also, @hongymagic could you merge develop? There's a merge conflict and I'm actually not sure how to resolve it. It was from my fix here: ae4d39ffef9b739e6f262c05c879af62b287ca98

OscarGodson:
Also, to be more clear about copying an pasting, I don't mind what HTML
does or doesn't get copied over. I just want copy paste to work like it
does now where new lines are preserved.

johnmdonahue:
@hongymagic I am still seeing the same issues with the editor including the original error and copy/paste stuff. Here's a video of what I'm seeing (http://f.cl.ly/items/1m0f1t232w1u0z3z122u/EE-feature-newline-issues.m4v). 

```
Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 epiceditor.js:234
_setSelectionRange epiceditor.js:234
shortcutUpHandler epiceditor.js:1032
EpicEditor.load.window.addEventListener._applyStyles.width epiceditor.js:1049
```

**Note:** most of the inserts are created just by hitting return - not from pasting - and the very last insert of text was a duplication of several lines never copied from the editor and inserted by just hitting the return key. 

Honestly, I'm not able to use the editor in this state and I can't imagine this being release-ready, if that was the intent. If the intent is to merge this in and address these issues on a separate branch, my only concern is that the resolution may not pan out in a timely enough manner and we end up with a main branch that includes code which we can no longer work around. Just my 2 cents.

/cc @OscarGodson

OscarGodson:
@johnmdonahue Yes, of course. This is a huge pull request, so we'll need to make it act, to the user, the same as it does on master.

@hongymagic Maybe you could make another pull request that _just_ has the 2 API methods you added, make sure thats bug free, merge that in, then we can do another pull request to do what this original ticket was originally for?

hongymagic:
@johnmdonahue paste has issues with this change. And that is why I suggested we move this off to a different branch and complete the code to stop browser's default behaviours when pasting, undoing and etc. I'll the branch on my fork and see if this will ever fly off.

@OscarGodson I think that's a better idea. This pull-request patches the way [contenteditable] works without providing alternatives for other useful methods such as undo, paste and whatever commands we've yet to test/discover.

I'll close this PR and think of another way to provide text manipulation.

johnmdonahue:
@OscarGodson: I'm sure you had your reasons for switching, but wasn't the editor originally using `textarea`? Would switching back help here at all? Seems like that would solve HTML copy/paste at the least, and I'm pretty sure you can hook into selected content, tabs etc as well no?

OscarGodson:
Switching to a textarea would make our lives a far bigger hell moving
forward. We can't expand into any new features without a content editable
element. Also, you might not remember that far back but there were lots of
styling issues.

hongymagic:
@johnmdonahue you can still use Range/Selection API for `textarea`. I wasn't aware that EE was based on `textarea` before, but the reason why one would use `contenteditable` is to render HTML as well.

hongymagic:
Quick mock of toolbar and EpicEditor interaction: [demo here](http://epiceditor-demos.herokuapp.com/) and [source here](https://github.com/hongymagic/EpicEditor-demos).

OscarGodson:
@hongymagic That works and looks great! I want to keep the buttons outside of the core code (just like you did), so in your pull request just have the API methods, tests, and docs. I'd love to integrate that demo you have into our core docs too since so many people are wondering how to do exactly what you just did. Super excited for this pull request :D What are your thoughts @johnmdonahue?`

hongymagic:
FYI: That demo is based on develop branch on your repo! I'm still thinking about how best to put that into API.

OscarGodson:
Oh, nice! I just checked out the code more in depth.

PR #169, opened 2012-08-11
Mocha Port

johnmdonahue:
## PR to get the conversation started.

This is basically just the first pass on the mocha port. It definitely needs some QA and I'd like to know how performance is affected on other machines.

A few things to note:
1. You can run a specific test block by clicking its header. You'll get something like this: `http://localhost:8000/test/tests.html?grep=^\.edit\(\)`, which is handy if you're only trying to test changes to a specific method. This is also why I changed from `index.html` to `tests.html` (as is the convention in mocha's own example) because grep won't work on an empty path: i.e. `http://localhost:8000/test/?grep=^\.edit\(\)`
2. Some test have been removed as redundant - these are mainly tests that checked state before running a test that depended on it but the methods used there should be tested in their specific API block anyway.
3. I reworded all "`should`" descriptions to hopefully better describe the test and fit the standard `expect` conventions. 
4. I used EE API calls where possible to hopefully increase performance and readability. This should be ok in most cases since they should be tested on their own regardless. (There are definitely still some slow tests that can be optimized though)
5. All tests have been broken into their own method specific files. This is handy to help get a grasp on coverage and hopefully reduce redundancy and improve findability.
6. I left the foounit test suite intact for now so that performance can be tested. I will do a quick rewrite of `jake test` task soon to just spin up a simple node server.

For anyone interested in playing around, `jake test` will still work to spin up a server until foounit is removed or you can just spin up any local server in the EE repo root and then hit localhost:8000/test/tests.html in a browser, e.g.:

```
$ python -mSimpleHTTPServer
```

I'll work on the larger `package.json`/`Jakefile` refactor on a separate branch when I get some more time. For now I think this works and in fact having Mocha deps included in the repo will allow us to hit the tests live on GHPages as we can now: http://epiceditor.com/spec/runner.html

Let me know any thoughts. :D


OscarGodson:
Totally didnt think about how we could test this on epiceditor.com. Thats pretty cool! I'll pull this right now and run the tests

OscarGodson:
Wow... what the. 2 seconds? How is it so fast?! Are all the tests actually in there?!

johnmdonahue:
Ha. You have a fucking monster of a machine. I don't see that kind of gain, but yeah they're all there except for the few I pulled. I tried to speed things up, good to hear you're seeing it. :)

OscarGodson:
There's something majorly different. It's 7 seconds cached with foounit. 2 seconds or less (best 1.89 now) with Mocha.

johnmdonahue:
Waiting for the laundry to dry so I decided to just go ahead tidy things up a bit. I got the tests to pass lint, added a simple node server to run them, and set up the Jakefile to use Mocha. The tests should probably be looked over still and definitely checked across browsers. But once that's done, I don't have much more to add.

Note: `jake test` will now spawn a simple node server on `:8888` (actually seems faster too)

OscarGodson:
Getting a "404 Not Found" going to http://localhost:8888/tests/tests.html (and I tried my IP)

johnmdonahue:
http://localhost:8888/test/tests.html

OscarGodson:
Noticed this too in Firefox. Not sure wtf it means:

OscarGodson:
The IEs are also completely failing because there's no `<!DOCTYPE html>` and trying to run in quirks mode. Also, none of the CSS is loading because of a MIME type mismatch? Are you trying to load CSS as raw text or something?

The tests are pass if I force it into IE standards and not quirks tho, but the page also looks pretty borked without CSS.

johnmdonahue:
That FF bug is weird. The leaked global is definitely not from EE tests and I don't see it in FF locally. OSX or Windows? Also could it be a plugin interfering?

OscarGodson:
OSX and I see it every time firebug is on. Dammit. I think that's probably it. Wonder why... It's the only plugin I have on Firefox.

It'd be nice to figure out why because I assume a lot of developers are going to have Firebug installed on their Firefox, but if not that's fine. Maybe put a thing saying to deactivate Firebug for the tests?

Firefox, Chrome, Safari, and IE9 all pass :thumbsup:

Just some minor tweaks and I'll merge it in for the 0.1.2 release.

johnmdonahue:
I went ahead and ported the Jakefile here since the `package.json` was needed to get the mime type fix. This removes the marked submodule as well as mocha, which means that test will not run on GHPages. If this is an issue we could add something to the build process that just copies the mocha files out of the module. To get this working, you should be able to just do `npm install`. A few notes:
1. The `package.json` needs to be updated with your email and any other info you want to add there.
2. If you want to publish this package to npm, I think it's ready. This would allow users to do `npm install epiceditor`.
3. The only dependency is now Jake, so the contributions guide needs to be updated as well - I'll try to get to that soon.

Let me know if you see any issues.

johnmdonahue:
@OscarGodson: I was unable to replicate the Firebug issue. Let me know if you're still seeing it with the latest commits.

OscarGodson:
Woooooooowwwwwww. I did an npm install to get it updated and there was a shit load of dependencies. Like, 100. Is that right?

And yes, I still see that error unless I right click on the Firebug icon -> disable all panels. Make sure you have Firefox 14.0.1 and latest firebug. Happens every time when the panel is either open or running in the background.

OscarGodson:
Also, just curious, is NPM the right place for something completely front-end?

johnmdonahue:
Hey so I just tried to install and get things running on windows... looks like the Jakefile will need to be updated. :)

The EE dependency list is short but each module has it's own dependencies. I don't really think it's a big deal personally but we can always revert to submodules if you want. 

Also, I upgraded everything to the versions you mentioned and still cannot replicate the Firebug issue.

Firefox 14.0.1
Firebug 1.10.2

johnmdonahue:
Yeah, npm can be for front-end code (e.g. ender modules, backbone etc.). Doesn't really need to be but just a thought.

OscarGodson:
No, not a big deal, just checking! I'll have @techwraith check. If he can't replicate I say fuck it, otherwise we'll need to figure it out.

OscarGodson:
So, @techwraith, if you get time can you just, with Firebug running in Firefox:

``` bash
$ git clone git@github:OscarGodson/EpicEditor
$ git fetch
$ git checkout mocha-port
$ npm install
$ jake test
```

That'd be a huge help.

zacparker:
Followed steps to repeat, all tests passing: http://cloud.zacparker.com/image/2m182r3m2P1o

OscarGodson:
Must just be my Firebug then. Fuck it. However, IE9 i'm getting global leak detected errors everywhere localStorage is called now, @johnmdonahue,

johnmdonahue:
@OscarGodson: Yeah I see that in IE as well and am debugging. I'm also seeing the global leak in FF for `getInterface` _without_ Firebug on Windows but still not on OSX. I have seen several reports of global leak detection issues on mocha and am working on a solution. Hang tight. :)

johnmdonahue:
@OscarGodson: I think ignoring leaks should be fine for now. If I get more time, I'll look into it more, but this seems more like an issue with IE localStorage and Mocha than EE itself. Thoughts?

OscarGodson:
Yes, it looks like some kind of bug with Mocha, but we'll need the tests to run in IE. IE9 and Firefox both pass now without errors!!!

In IE9 there's a weird bug where none of the CSS loads or something until the tests are done. Pretty minor, but if you had any ideas let me know.

Anyways... :shipit:

johnmdonahue:
We could maybe delay the tests, but IMO it's not really worth it. The style stuff is just flare anyway :sparkles: :)

OscarGodson:
Yep! Agreed. I'll let you do the honors of clicking the shiny green button. :)

johnmdonahue:
One last thing... @zacparker: thanks for running the tests! :)

zacparker:
no problem! :)

PR #175, opened 2012-08-16
Added Bartik theme

fuddl:


OscarGodson:
This needs a background color on the body. :) I'm not sure exactly what color Drupal's WYSIWYG background is.

OscarGodson:
Also this is what code blocks look like. Is this right? Doesn't appear to be, but I've never used Drupal.



I'm really liking the theme tho. Maybe this could be the default? @johnmdonahue? Here's what it looks like with a `#fcfcfc` background:

fuddl:
I added a white background.

and btw this is what a fresh drupal looks like:

Bartik

OscarGodson:
Did you see my message above about the code block colors?

fuddl:
is it wrapped in <code>pre</code> and <code>code</code>?
like:

```
<pre><code>
if (foo) {
  alert('test');
}
</code></pre>
```

OscarGodson:
Yes. It matches the Google and GitHub style of formatting code. Personally I think it's kinda stupid to wrap it like that, but it's the standard.

OscarGodson:
Merged. Thanks man! I think 3 preview themes makes it enough to have some previews too, so I'm going to make a ticket to show themes in the README/docs.

PR #187, opened 2012-10-23
EpicEditor with google-code-prettify

if1live:
Hi. I am a developer who interesting in this project.
This is my first commit to someone's opensource project, so I don't know how to write pull request.

I combine EpicEditor with google-code-prettify. In preview, code highlighting work.
# Fixed File
- main.js
  - Add some code for test
- epiceditor_prettify.js 
  - modify google-code-prettify to combine
- epiceditor_prettify.js
  - base css of google-code-prettify
- epiceditor.js
  - Fix preview function, code for "<pre.>" and "var opts"


OscarGodson:
This is great! This is something we've been wanting to do for awhile now. So there's some things left tho:
1. It looks like you used tabs instead of spaces? If you look at the options object in the diffs above you should notice that `code` for example doesn't line up. If you look at the [raw file](https://raw.github.com/if1live/EpicEditor/e6a2bde529815a8d4337c9433deece22ed48b4f9/README.md) notice `code` is way more indented then the rest of the file. This is probably because you didn't run this through JSHint which leads me to #2 :)
2. You need to build these files with `jake build`. Currently you did all your work in the built files so the next time anyone makes a change all your changes will get overwritten. You need to make your JS changes to `/src/editor.js` then run `jake build`. This will create the minified and un-compressed version as well as lint your code. For your README.md changes, make sure you run `jake docs`. Our index.html files is dynamically generated from the README.md file. Take a look at here under "Requirements" and below: https://github.com/OscarGodson/EpicEditor/wiki/Contributing -- If you have any questions about this feel free to ask!
3. The way you implemented this is great. You add another script tag. This keeps the main file small for people who don't need it, but does EpicEditor work, without errors, if you _don't_ put in the prettify script? It should work without prettify since this is more of an add-on.
4. I'm thinking all the prettify stuff should go in /epiceditor/addons. As I said in #3, this makes it so people can simply delete whichever addons they dont want. I'm not sure how addons should work tho. Maybe there's an addons array in the options you can list addons you want? Thoughts? Ideas?

Let me know your thoughts and thanks a lot! This is a highly requested feature.

if1live:
Contribting document! I read it and apply to next commit.
addon is good idea. i'll try it.

OscarGodson:
:+1:

PR #188, opened 2012-10-24
Feature/codepreitty Syntax Highlighting in preview mode

if1live:
Contributing document based pull request (I forgot writing commit messge with ticket number.sorry)

customized google-code-prettify addon work with jake except lint. (google-code-prettify is not JSHint friendly code. To pass JSHint, Fix ALL google-code-prettify code)

If include epiceditor.codeprettify.js before epiceditor.js, Syntax highlighting in preview mode is enabled.
else, It work as original EpicEditor.


OscarGodson:
It's looking good, but I think the addons shouldn't need to be hardcoded in `editor.js`. Basically, there should be a convention where you can do something like this when setting up EE:

``` js
opts = {
  addons: [
    'google-code-prettify'
  ]
}
```

And it'll go look for `/addons/google-code-prettify/` if it's there it'll look for a sort of manifest file. Maybe something  like:

**manifest.json**

``` json
  "js": [
    "javascript-file-1.js",
    "javascript-file-2.js"
  ],
  "css": [
    "style.css",
    "other-styles.css"
  ]
```

If not, basically every addon will have to add a bunch of if/else statements in the editor and it'll keep growing. This way it's one block of code that checks dynamically. 

Or, I like what you did in the Jakefile. Maybe there's a way to pass in addons to Jake and have it build in all the addons.

What are your thoughts on this @if1live?

@johnmdonahue - I'd love to hear your ideas and suggestions :)

OscarGodson:
Oh, and you don't need to open a new pull request each time. You can just keep committing to your fork and the commits will keep coming into this pull request as long as you commit to the same branch.

if1live:
Manifest is good idea. I'll try it.

I don't know correct term in jake. I say editor, utils, docs as task. Next commit, I will create new task, addons.
On running build:addons, lint addons file And minify addon js file. Run build task mean build:editor and build:addons.
Next step, try searching all addons and build automatic, without add new addon inside jake file.

To run addons dynamically, Need function like postPreview(something like trigger). If preview function called, execute prePreview and original Preview and postPreview. addons run in pre/post function. Manifest+trigger will remove hardcoded addon code.

if1live:
add some javascript file into src/, and `jake lint` or `jake build`. This task will work well.

if1live:
I try dynamic load/build addons with manifest.
I failed making manifest as pure json. (I can't find how to include json file as JS variable.) I make manifest as global variable. If you know how to load JSON manifest file as JS variable, could you give me reference/etc? 

``` js
EE_addon_manifest = {
  "google-code-prettify" : {
    "js": [ "google-code-prettify.js"],
    "css": [ "prettify.css" ]
  }
};
```

Predefined CSS, JS are attached to previewer iframe. With manifest, this code have extensibility.

``` js
if (typeof EE_addon_manifest != 'undefined' && EE_addon_manifest) {
      for (var j = 0 ; j < self.settings.addons.length ; ++j) {
        var addonName = self.settings.addons[j];
        var manifest = EE_addon_manifest[addonName];
        for (var x = 0 ; x < manifest.css.length ; ++x) {
          var cssFile = manifest.css[x];
          var cssPath = self.settings.basePath + '/addons/' + addonName + '/' + cssFile;
          _insertCSSLink(cssPath, self.previewerIframeDocument, '');
        }
        // for javascript. almost same code....
      }
    }
```

Addons are attached to iframe. Addon needs to access editor. So, add variable new variable 'editor' to iframe document.

``` js
self.previewerIframeDocument.editor = self;
self.editorIframeDocument.editor = self;
```

In addon, register addon function to editor trigger.

```
// Register prettify function to trigger list
  document.editor.on('preview', prettyPrint);
```

At Last, If you press preview button, Trigger function list will fire!

jake build is work well. But, I can't make decision where to put addon file, manifest file, build destination path, etc. I don't want to mess up source tree. Please check source path, dest path,etc.

P.S. Not documented yet. 
P.S. My english skill is READ-ONLY. If you can't understand what i'm saying, please read code. Maybe code is easier than this discussion.

OscarGodson:
Wow, lots of work here. Give me a little time to dig through read and test it! Nice work!

johnmdonahue:
Hello! Apologies for the delay.

There's a lot of good work here @if1live !! 

I may be missing some of the nuance of what's being addressed here, but my immediate concern is that adding plugins to `/src` will not be a manageable strategy. Plugins/add-ons should remain for the user to add rather than bundling with EE in the same way that EE can be extended to use any parser either [during the build step](https://github.com/if1live/EpicEditor/blob/feature/codepreitty/Jakefile.js#L164) or [passing to opts](https://github.com/if1live/EpicEditor/blob/feature/codepreitty/src/editor.js#L351).

I'm all in favor of using the opts method to pass addons as @OscarGodson mentioned. It's simpler than the build method. The reason for the parser build is that EE **must** have a default parser but if someone really wanted to optimize EE they could replace the bundled parser with one of their choosing. This is not the case for add-ons.

Personally I think a manifest can be avoided as well. Correct me if I'm wrong, but this could all be handled by the assets options as discussed in #142. If something needs to be executed on preview, as would be the case with prettify, the user could just listen for that event, or in fullscreen listen for update maybe. You could load any assets you want to the frames, then hook into event handlers for the rest. Does this make sense? Maybe all we need to do here is build a proof of concept using the opts method and then provide links to example repos in the docs of how Google Prettify or other add-ons can be implemented.

OscarGodson:
No update. Closing. Feel free to reopen.

PR #190, opened 2012-11-15
Ticket #4 - add DOCTYPEs to the iframes.

jeremybenaim:
Not sure it'll fix display bugs but whatever. ^^


OscarGodson:
How does the browser know?! It's like magic. OK, I'm pull in your commit and see if IE looks any different! Thanks. I did not expect it to know to put the template in the BODY.

OscarGodson:
Well, looks like IE8 is rendering the DOCTYPE and everything seems correct. Need to test the rendering stuff more now. 

HEY LOOK. You'll be a contributor now! :smiley:

OscarGodson:
Just for records sake. I was able to revert my CSS hacks AND it fixes 99% of the rendering issues. 

https://github.com/OscarGodson/EpicEditor/commit/0fb3c56abb8727fd916b7e94026d27c1dafaae15

Issue #22, opened 2012-02-21
Get fullscreen working correctly in Firefox


OscarGodson:
It's totally fucked right now

OscarGodson:
It appears to be an issue with the API wrapper we're using: http://johndyer.name/native-fullscreen-javascript-api-plus-jquery-plugin/

The `fsElement.addEventListener(fullScreenApi.fullScreenEventName,function(){` is never getting called, therefore, fullscreen never really gets setup

OscarGodson:
Fullscreen is unsupported if you are using alphanumeric keys, well, it'll work you just get a massive PRESS ESC TO EXIT FULLSCREEN MODE constantly. I'm closing this and adding this to the list of unsupported browsers until Firefox gets its shit together.

OscarGodson:
The creator of the wrapper pointed out FF needs to have the event attached to the document, not the element: https://gist.github.com/9ace07aa0c7e532f9311

PR #202, opened 2013-01-18
Native fullscreen support in Firefox and Opera

nkovacs:
I had to resize the elements in the resize event listener even when using native fullscreen, because otherwise in opera (at least on linux) the editor would have the height of the document instead of spanning the whole screen (probably a bug in opera). This exposed another bug: _exitFullscreen calls the native cancel/exit function, which in turn calls _exitFullscreen again.

### DIFF
--- a/epiceditor/js/epiceditor.js
+++ b/epiceditor/js/epiceditor.js
@@ -459,6 +459,9 @@
       , _elementStates
       , _isInEdit
       , nativeFs = false
+      , nativeFsWebkit = false
+      , nativeFsMoz = false
+      , nativeFsW3C = false
       , fsElement
       , isMod = false
       , isCtrl = false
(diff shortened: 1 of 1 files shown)


OscarGodson:
This is super great and I've been wanting to do this myself, but haven't had time to do it yet. The code for the most part looks good.

Make sure to modify `src/editor.js` then run `jake build`. This will create a minified version as well as an uncompressed version. It will also lint the code. Here's some docs: https://github.com/OscarGodson/EpicEditor/wiki/Contributing (don't worry about most of it. Just install the right stuff and then go to "Flow" step 5ish).

The only bug I found was in Chrome when I open fullscreen and close it the preview icon is wrong and shows that im in preview mode when im in edit. Clicking it gets it into a funky state. If you can't repro on your end I can do a screencast if you want.

As for Opera, we don't support it. You can take a look at Opera support if you want in another PR but the markdown conversion doesn't even work. It works in IE8 even, but not in Opera so I've basically just opted out of Opera support. It's good to have w3c version of Fullscreen so when Moz and WebKit switch it works, but don't worry about getting Opera right and don't do any gnarly hacks just for Opera since the editor already doesn't work right in it.

Again, huuuuuuuge thanks. You rock. If you could just fix the Chrome bug and put your code in src/editor.js and run jake build I'm pretty sure I can pull it in. Make sure the tests pass as well :)

:clap:

OscarGodson:
Just for records sake and searchability: this will _actually_ fix #22 now that Firefox has implemented fullscreen correctly.

nkovacs:
Opera implements the W3C standard, that's why I used it to test. The resize event handler shouldn't break anything in other browsers.

I can reproduce the bug with the preview icon, I'll try to fix it.

PR #203, opened 2013-01-18
Ticket #22 - native fullscreen support for firefox and w3c standard

nkovacs:
Forked it into a separate branch, so I could do it more cleanly (modified src/editor.js first, then jake build), and also fixed the preview icon bug that I introduced. All tests passed

https://github.com/OscarGodson/EpicEditor/pull/202


OscarGodson:
This is great. Let me play with it more later and I'll let you know. Thanks again!

OscarGodson:
This looks good to me. I'll pull it in. I'm going to change all your:

``` js
if {

} else if {

}
```

to

``` js
if {
}
else if {

}
```

to match the rest of our code. Not sure if you want to do that tho so there's a better history on those lines or not, but if you do let me know and I'll let you change those and then ill merge it in.

Something else to think about and I'm not sure if it's a problem or not, but what happens when Chrome/FF supports the W3C method. Do they drop the webkit fullscreen stuff or support both? If they support both what happens with our code? Not that you would know, but if you did I'd love to hear any insight you have.

nkovacs:
Feel free to change the if blocks. I doubt they'll keep the prefixed versions once they adopt the w3c standard (which is still just a draft), but if they do, only the prefixed code will run, because of the `else if`

Issue #204, opened 2013-02-01
&nbsp; displays with Firefox and getFile .content


zspecza:
Playing around with 0.2.1 along with syncing with a `<textarea>` and syntax highlighting via highlight.js, saving content to a MongoDB database, I find that within code tags, any indentation via spaces is outputted as `&nbsp;` for each space except the last, that one outputs fine. 

I've tried breaking things a little by removing the `&nbsp;`'s from `.replace()` in lines 183, 1377 and 1380 of epiceditor.js, but I still see that on output even if I create a new document after the changes... Is really doing my head in. I've also tried looking at highlight.js but there's not even a mention of `&nbsp;` there that I can find with Sublime Text.

This _only_ happens in code tags, nowhere else... Any thoughts?

OscarGodson:
I'm not sure off the top of my head, but I will look into it. I believe highlight.js actually adds those in their code and I'm not sure the sequence of events that happen. However, are you using the textarea option I just added? And does this happen without highlight.js but just manually typing in:

```
<code>
Foo
  Hello
        World
</code>
```

? Also, it'd be awesome if you could upload a testcase or put something on JSBin :)

zspecza:
I am using the textarea option you just added. It happens without highlight.js too. http://jsbin.com/udaboy/1/edit
Seems it works on non-code tags too. That didn't happen on my local instance after converting to html, which was done simply by wrapping `self.getFiles(textareaFileName, true).content` in a `marked()` call on lines 931 and 935.

OscarGodson:
Perfect! Thanks man, I'll take a look at it tonight or tomorrow. This helps a lot.

OscarGodson:
Hmm, ok, having trouble reproing. I put this in the editor. All export functions and the text too doesnt have any `&nbsp;`. Do you have an example of the exact steps or a screencast to repro?

![Screen Shot 2013-01-31 at 5 49 52 PM](https://f.cloud.github.com/assets/20615/116966/dd2ef98e-6c11-11e2-90d3-43ce9060841e.png)

zspecza:
That's strange. Hmmm... 

Have you looked at the JSBin I provided? That reproduces the issue as far as text goes. Would the browser I'm using make any difference? Using the latest Firefox.

OscarGodson:
That could be it. I tried chrome. The screenshot is from your JSBin.

zspecza:
[...]

PR #206, opened 2013-02-07
Ticket #204 - sync textarea with exported text rather than raw text

lamplightdev:
As per ticket #204 a synced textarea was giving `&nbsp;` in place of regular spaces on Firefox. This was because the textarea was being synced from `getFiles` which gives the raw content from localStorage. This doesn't allow for the vagaries of space handling in contenteditables accross browsers. `exportFile` does aleady allow for these inconsistencies however so I have changed the code so that textareas are now synced using `exportFile` rather than `getFiles`.


OscarGodson:
This patch does in fact work. `getFiles` _should_ work the same. I've been banging my head trying to come up with a solution to get `getFiles` to work correctly (`_setText` and `_getText` need some refactoring), but I guess we can do `exportFile` so that at least develop works for now.

Couple requests for the pull request :)
1. Could you send a patch that doesn't have so many whitespace changes?
2. Can you build your change? (`jake build`)
3. Shouldn't break any, but make sure you run the tests also (`jake test`)

Thanks!

lamplightdev:
That should be sorted now, although I made another pull request with the changes in.

PR #207, opened 2013-02-07
Ticket #204 - sync textarea with exported text rather than raw text

lamplightdev:
As per ticket #204 a synced textarea was giving `&nbsp;` in place of regular spaces on Firefox. This was because the textarea was being synced from `getFiles` which gives the raw content from localStorage. This doesn't allow for the vagaries of space handling in contenteditables accross browsers. `exportFile` does already allow for these inconsistencies however so I have changed the code so that textareas are now synced using `exportFile` rather than `getFiles`.


OscarGodson:
You rock, man!

lamplightdev:
keep on rockin' in the free world :)

PR #208, opened 2013-02-09
Update src/editor.js

ghost:
OCD Comment Fix


OscarGodson:
Good catch! Thanks. Your comment would have made a better commit message than the current commit message tho :P

PR #209, opened 2013-02-09
Add load() to the `new EpicEditor(opts)`

ghost:
Just pointing this line in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ var opts = {
     preview: 80
   }
 }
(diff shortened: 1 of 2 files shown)


OscarGodson:
Its purposely not in the options section because we keep each example as minimal as possible to prevent confusion. In this case you may not want to load the editor yet :)

ghost:
OK :) Thanks for your quick response anyways.

PR #214, opened 2013-03-05
Enabled fullscreenEnabled configuration setting, which allows you to disable the full screen button if desired.

dobozysaurus:
This patch request allows a user to disable the full screen button if they desire via an additional configuration setting (fullscreenEnabled: false). The full screen button remains enabled by default.


OscarGodson:
This is awesome. It's something I've been meaning to do. Could you make it a little more broad though? By that I mean, allowing you to disable any button rather than just fullscreen? I've had about 3 or 4 requests asking how to turn off the preview button. I was thinking something like:

``` js
// Turn off the button bar altogether 
var opts = {
  button: false
}
```

``` js
// Turn off fullscreen
var opts = {
  button: {
    fullscreen: false
  }
}
```

``` js
// Turn off preview
var opts = {
  button: {
    preview: false
  }
}
```

Also, feel free to build it (`jake build`). Lastly, I'll accept the PR without tests, but before I can pull it into develop I'll need tests. If you can add them that'd be sweet :) otherwise I'll do it when I get some time.

dobozysaurus:
There you go - added support for hiding the preview, fullscreen, and edit buttons. The modified code now supports buttons: false (all disabled), buttons: true (all enabled), as well as stuff like buttons: { preview: false } (disables preview, leaves other buttons enabled).

dobozysaurus:
Ah, and I added some tests for the button behavior as well.

OscarGodson:
Wow, looks perfect! I'll test it out soon. Thanks a lot!

OscarGodson:
Thanks again. This looks great. 2 minor things:
1. We decided to singularize our options (i.e. `theme` and `shortcut`) so it makes sense reading it like `opts.theme.preview`. Could you just change `buttons` to `button`?
2. You can delete your checks for the private properties in your tests. We try to avoid testing private things because they're meant to change. We're more worried about breaking API changes, so the other tests you have there checking the length of the buttons works fine.

After that I'll merge this in. HUGE THANK YOU AGAIN!

dobozysaurus:
No problem; happy to help. I've updated the tests and singularized the option.

PR #220, opened 2013-04-08
Add .editorconfig to make it easy for developers to follow code style

hongymagic:
If it were upto me, local `.vimrc` file would suffice, but `.editorconfig` is supported in more editors. I personally like tabs for indentation, and all my editors are configured to behave this way. Using `.editorconfig`, I won't have to worry about code style.

I think this is a easy win :smile:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+; 2 spaces for indentation for all JavaScript files
+[*.js]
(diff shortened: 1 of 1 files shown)


OscarGodson:
Sweet, and this works in Vim?

hongymagic:
Yep absolutely: http://editorconfig.org/#download

PR #232, opened 2013-05-29
added the ability to wrap a textarea

bytespider:
Wrap a textarea with the container and use that as the sync textarea


OscarGodson:
So, to be clear, if you had:

``` js
<div id="epiceditor">
  <textarea></textarea>
</div>
```

It would use that textarea, automatically, to sync content?

bytespider:
That's the idea

bytespider:
Removed check and IE10 tested

OscarGodson:
Rad! I'll check play with this when I get home tonight then. Thanks, man!

OscarGodson:
Pulled this down and it doesn't appear to work for me. 

``` html
  <div id="epiceditor">
    <textarea></textarea>
  </div>
  <script src="/epiceditor/js/epiceditor.js"></script>
  <script>
  </script>
```

The textarea appears to be removed. This line is doing it (99.9% sure)

https://github.com/OscarGodson/EpicEditor/blob/develop/src/editor.js#L532

Actually, I'm not sure how this worked for you since `self.element` is whatever you give it: https://github.com/OscarGodson/EpicEditor/blob/develop/src/editor.js#L375-L380 :P

bytespider:
The loading from the contained textarea definatly works for me, however what I hadn't tested or checked was that the textarea is placed back on the form after the editor is done.

Think I might need to play with this a little before you consider pulling this.

OscarGodson:
oh, I see. This was just for initial loading. Yeah, most people requested the textarea feature so that they could submit a form the old school way so the textarea has to be in the DOM when the form is submitted.

bytespider:
What's your feelings on CSS added via JS? In this case,`'self.settings.textarea.style.display = 'none';`?

OscarGodson:
As long as the docs make it clear that by putting the textarea in the epiceditor container will hide the textarea and replace it with epiceditor.

bytespider:
I believe this is complete. Fancy giving it a go?

OscarGodson:
Will do, sorry for the delay. I've been reviewing PRs for the 0.2.1 release. I'm going to package and ship 0.2.1 first and then I'll take a look at this. I will try to package and ship it tonight and take a look at this tomorrow. Thanks again!

bytespider:
Ah, I thought perhaps you weren't sure I'd finihed making changes.
—

OscarGodson:
OK, 0.2.1 was just released. I'll take a look at this ASAP (tonight/tomorrow) and test it in all the browsers.

OscarGodson:
OK, checked and everything looks and works good. Few things:
1. Can you use `_applyAttrs`?
2. Can you do `typeof x` instead of `typeof(x)` as it fits more inline with the styles
3. Add docs (edit the README.md the file and then run `jake docs`)

If you don't have time or don't want to I can take care of these myself.

OscarGodson:
P.S. sorry it took so long to get to this. Was busy with stuff :(

OscarGodson:
bump?

bytespider:
Yeah sorry, newborns can be a little distracting. I'll look into this weds/thurs as I'm back to work then
—

OscarGodson:
@bytespider totally understood. Thanks :)

OscarGodson:
I just noticed you put some commits in. Is that everything you wanted to get in?

bytespider:
I think so, unless you had some comments on the docs?
—

OscarGodson:
Rad, no, I just never got a comment from you. I don't get notified of commits, just comments :) I'll look at it during lunch

bytespider:
Oh right, Thought it did as the commits were inline on GH
—

OscarGodson:
When I mentioned applyAttrs, I didn't mean that one style, I meant where you set that block of attrs on L540-L543.

bytespider:
Clearly I miss understood you. Will sort.
—

bytespider:
Okay. So I used _applyAttrs, however it refuses to set frameborder. 
Therefore I updated _applyAttrs to use setAttribute instead.

OscarGodson:
I will review this ASAP :) I wanna get this into the next release!

PR #235, opened 2013-06-04
Autosave/save split

Gankra:
Tried to conform to your style as best I could. Hopefully I didn't miss anything. Not clear if this change merits a new test or not.

Also slipped in a fix to your test code to make the linter stop yelling at me when I ran `jake test`


OscarGodson:
Could you just assume if it's a preview draft that its an internal save therefore _is_ an autosave?

And yeah, need a test that at least checks that autosave does not fire save event, but an autosave event and that save fires a save event and not autosave. Since these are private params you don't need tests for those, but for the events yeah. 

_edit: said "not an autosave", meant is an autosave if it's a preview draft._

Gankra:
Okay, how's that?

OscarGodson:
I think it's good. I'll play with it tonight. This is great. Thanks for being so responsive :)

OscarGodson:
So, playing with it I found a slight problem with our logic. Internal saves to the "draft" file storage arent technically autosaves at all because they are stored in memory and are always lost between page refreshes. An example:

``` js
    var editor = new EpicEditor({file: {autoSave: false}});
    var autosaves = 0;
    var saves = 0;
    editor.on('autosave', function () {
      console.log('Autosaves: ', ++autosaves);
    });
    editor.on('save', function () {
      console.log('Saves: ', ++saves);
    });
    document.querySelector('.save-btn').addEventListener('click', function () {
      editor.save();
    });
    editor.load();
```

If you then type some stuff, and click preview, you'll see an autosave. If you refresh the page tho, your change is lost. Autosave should probably only mean autosaves on that timer. As you also mentioned in that ticket it should not autosave if the content isn't different. We don't have to do that part in this PR (unless you wanted to), but I think that''s what would make the autosave event handy (i.e. when did autosave catch a change).

Gankra:
Okay, so revert back to only considering it an autosave when the flag in the method is explicitly set, and only explicitly set it when both:
- It is being called by the autosave timer
- The save actually changes the stored data

Do you have built in functionality to determine whether an actual change has occurred other than just oldText==newText? It looks like the "input" event introduced in HTML5 is what we would want to listen for. Although it'll fire even if they edit the text in a way that results in the text not actually changing (such as CTRL+A, CTRL+C, CTRL+V), but that doesn't really matter that much, as far as I'm concerned. So just run the autosave timer, and if an input event was fired in the last interval, it autosaves?

OscarGodson:
Thatd rock. Sorry for incorrect info :\ no, I just check if the text is different. If you have other suggestions, go ahead!

Gankra:
I'll toy around with some ideas in the morning.

Gankra:
I'm not entirely clear on how you want to count the two saves done in .open and .importFile. They aren't in-memory, they aren't part of the autosave routine, and they aren't explicitly triggered by outside action.

OscarGodson:
What do you think? I'm leaning towards autosave since it was automatically saved.

Gankra:
Doing an explicit text comparison during saves doesn't seems to not be any kind of bottleneck in firefox/chrome (works fine @ 60 000 chars). I'm not inclined to change it unless it's a problem.

Gankra:
Alright, I used my best judgement on the .open and .importFile saves (assumed they're "hard" saves). Totally willing to consider them autosaves. Your call.

OscarGodson:
If you think save makes more sense, then sure. As long as the docs reflect this. Right now they say:

> Fires whenever the file is saved by EpicEditor automatically.

Something like this maybe:

> Fires whenever the autosave interval fires and the file contents have been updated

And for save:

> Fires whenever `save()` is called manually, or implicitly such as `importFile`, `open` {PUT OTHERS HERE}.

Gankra:
Done.

OscarGodson:
Can you run `jake docs` to generate a new index.html and just make sure it looks all ok as well? I'll check everything out on my end today as well.

Gankra:
Oh, whoops. Didn't know about that one. Fixed.

OscarGodson:
Ran tests in IE9, 10, Chrome, and FF. Manually played with it in all those browsers too. Looks good. Thanks!

OscarGodson:
@Gankro  I believe this commit broke `save()`. I was trying release 0.2.1 and going through all the tickets to QA them one last time and I ran into an issue (use Chrome or IE):
1. Turn off autoSave
2. Type some new stuff into the editor
3. Refresh. No changes were saved. Good!
4. Type more new stuff
5. Save manually with `editor.save()`
6. Refresh. No changes were saved. Bad :(

Mind taking a look? I can take a look this weekend if you can't get to this. If you find out this was you, mind writing a test so this doesn't regress?

Also, I said use Chrome or IE just because Firefox will store the content of a contenteditable on refresh like most browsers do with form inputs. This can make it sort of a pain when testing this.

Gankra:
I generally always test in chrome/ff/safari. Guess I wasn't paying attention...

Gankra:
Trying it out on Chrome Version 27.0.1453.110 on Mac OSX 10.8.3, I can't repro it on the branch from this pull request. So if this branch broke anything, it's in interaction with something else that was merged in. (Or it's some weird machine-specific thing). Can you direct me to the branch where you're seeing this behaviour?

Gankra:
For reference, this also isn't repro-ing for me on current develop. How are you disabling autosave? I'm applying this patch to do it:

```
-- a/docs/js/main.js
+++ b/docs/js/main.js
@@ -31,7 +31,7 @@ $(function () {

   var opts =
       { container: 'example-1'
-      , file: { defaultContent: "#EpicEditor\nThis is some default content. Go ahead, _change me_. " }
+      , file: { defaultContent: "#EpicEditor\nThis is some default content. Go ahead, _change me_. ", autoSave: false }
       , focusOnLoad: true
       }
     , editor = new EpicEditor(opts).load()
```

Gankra:
In IE10, Windows 7 Enterprise SP1, I get this bug in both develop and master. But that might be because I'm running it from file://?

Edit: saving doesn't work at all, auto or not, for my local IE in master or develop :S

Gankra:
Autosave works in my IE on epiceditor.com. Guess it's a file:// thing? Although I don't know how to disable autosave properly in epiceditor.com

OscarGodson:
You can turn off autoSave by doing `EpicEditor({ file: { autoSave: false } })`

However, I think I know the problem...

I had two tabs open. One of them still had autoSave enabled from testing something. This meant the old content was autosaving in tab 2 after I was doing editor.save() in tab 1.

OscarGodson:
So, I guess, not a bug. Sorry!

Gankra:
Yesssssss

PR #236, opened 2013-06-05
Ticket #204 - escaping content in firefox getFiles

Gankra:
This seems to make getFiles behave just like exportFiles wrt white space characters.


OscarGodson:
I'll need to test it, but looks good! I'm so glad this is most likely fixed.

Just a couple small things:
1. Can you add a test to make sure no `&nbsp;` get exported. This has regressed a couple times in the past.
2. Generally, we put any helper functions up top above the EpicEditor constructor prefixed with an underscore and methods are all below.
3. Could you add JSDocs for those two functions? Example: https://github.com/Gankro/EpicEditor/blob/fc271ff6d85761fdb0f4295d151f227535ab4368/epiceditor/js/epiceditor.js#L7-L12

Gankra:
Just noticed a completely insane bug this may cause in firefox:

Paste the sample text: 

```
foo
  foo
          foo
```

Pull up the DOM inspector and note the html in the contenteditable area (it's what you would expect).

save, refresh

The second and third line have now lost a space on the page. In the DOM inspector the correct number of spaces, however, _are_ there. Now repaste the sample text. The DOM doesn't change, but it now looks correct in the textfield.

WTF

OscarGodson:
Yep :( this is why I mentioned to make sure the correct amount of spaces stay there. I've also ran into this. The fact Firefox doesn't support innerText makes all this stuff a major pain in the ass. And, that one space can totally mess up the Markdown for things like code blocks where people indent 4 spaces that would then become three. Major WTF.

Gankra:
Perhaps, at this point, it's best to just establish that getFiles does not necessarily return a nice version of the text, but rather a version that the browser will "like"? If they want something useful it seems like exportFile is the way to go, no?

OscarGodson:
Hm. You know. That isn't a bad idea. The main problem is I needed a way to get info about the file like the modified date as well as add meta data. Here's an idea:

What if exportFile can export the JS object or JSON? So, if you export as text (null) or HTML it returns a string. If you export as `json` you'd get:

``` json
"{"example-1":{"content":"Hello world","created":"2013-03-08T07:31:31.035Z","modified":"2013-06-05T18:21:11.849Z"}}"
```

or `js`?

``` js
{"example-1":{"content":"Hello world","created":"2013-03-08T07:31:31.035Z","modified":"2013-06-05T18:21:11.849Z"}}
```

This would mean people would use getFiles a lot less for content and meta data. The only use case left would be to get all files. We could, technically, change getFiles to exportStorage() which is just a data dump like getFiles and we could make it more obvious that this is a data dump from the browser's localStorage.

What are your thoughts?

Gankra:
I think it would make sense to make `getFiles()` just report all known names with timestamps (so basically what it is now without the content attribute), and then if anyone is interested in the content they can make a separate call to `exportFile(someName, type)`, where excluding the type gives the raw format the browser would want (as it does now).

`getFiles(name)` would just give you metadata about that one file.

OscarGodson:
Cool, I like that idea. I'll make a different ticket for `.exportFile(name, 'json')`. Since we'll be removing `.content` I will need a way to easily merge the content and data into a JSON string to be sent to a server.

Gankra:
Is that what getFiles was for? Huh.

Whelp, I think I can close this down.

OscarGodson:
It was used for a few things (or this is what I've seen people use it for):
1. List all file names to make a file picker.
2. To be sent to the server as a JSON blob `JSON.stringify(editor.getFiles())` or `JSON.stringify(editor.getFiles(name))`
3. Check timestamps of the modified date so it could check the server if there was a newer copy on the server someone else had modified.
4. For getting content (which they should use exportFile for)

Gankra:
I really can't fathom wanting to send every single saved file ever to the server with wild abandon like that.

OscarGodson:
Yeah, probably not all, but the single file, yeah

Gankra:
While cleaning up the internal API calls to apply my new changes I stumbled across the following line: https://github.com/OscarGodson/EpicEditor/blob/develop/src/editor.js#L421

Surely the next line makes this line completely useless?

OscarGodson:
Yeah, that would just be setting it to undefined. Should be able to remove that. Good catch.

PR #238, opened 2013-06-05
Ticket #204 - refactoring getFiles

Gankra:
Not only did I refactor getFiles to match what we discussed, but I fiddled around with the firefox sanitization regexes in a way that I think that resolves the core issue.

The problem, it seems, was that you were only escaping _pairs_ of whitespace, which would either leave a space at the start of line (which firefox would eat), or would flub up when there was an odd number of spaces by having two spaces together at the end of the sequence (which firefox would eat). The secret sauce, then, was to first escape "space-space-space" into "nbsp-space-nbsp", and then apply the normal "even" rule. I also added a rule that first escapes spaces after new lines to ensure all lines start with a nbsp, if any space.


OscarGodson:
Sounds great. I'm at work, but tonight I can check it out in full. Only issue I see off the top of my head is in exportFile it says it might be browser specific. Does this mean exportFile will now return nbsp's? If so, that wont work since that at the very least has to return a raw string so people can use this to send to the server.

Gankra:
Basically I'm just saying "the default is just what the browser wants in a contenteditable". If they want to send something to the server, they should use the "text" type.

OscarGodson:
As a developer, would you ever want this? The default should still be clean text because I don't see why any dev would ever want the contenteditable content unless he was trying to do some weird editor display hacks. Generally people just do:

```
editor.exportFile()
```

This would break that functionality, right?

OscarGodson:
Awh, disregard. I see that `kind = kind || "text"` still. Then, cool, looks good :)

README.md
```diff
@@ -287,11 +287,11 @@ importFileBtn.onclick = function () {
 
 ### exportFile([_filename_],[_type_])
 
-Returns the raw content of the file by default, or if given a `type` will return the content converted into that type. If you leave both parameters `null` it will return the current document's raw content.
+Returns the raw content of the file by default (warning: this may be browser-specific), or if given a `type` will return the content converted into that type. If you leave both parameters `null` it will return the current document's raw content.
 
 ```javascript
 syncWithServerBtn.onclick = function () {
-  var theContent = editor.exportFile();
+  var theContent = editor.exportFile(null, 'text');
```
OscarGodson: This example isn't as nice as the simple editor.exportFile(). Since text is still the default there isn't any need to show this IMO.

README.md
```diff
@@ -287,11 +287,11 @@ importFileBtn.onclick = function () {
 
 ### exportFile([_filename_],[_type_])
 
-Returns the raw content of the file by default, or if given a `type` will return the content converted into that type. If you leave both parameters `null` it will return the current document's raw content.
+Returns the raw content of the file by default (warning: this may be browser-specific), or if given a `type` will return the content converted into that type. If you leave both parameters `null` it will return the current document's raw content.
```
OscarGodson: This part of the docs is a little scary:

> Returns the raw content of the file by default (warning: this may be browser-specific)

It makes it sound like if you just `exportFile()` you'll get a browser specific export, which isn't true. Also, I have no idea how I missed this, but the types aren't listed. No mention that it supports HTML output! 

Maybe this should read:

> Returns the plain text of the file by default, or if given a `type` will return the content in the specified type. If you leave both parameters `null` it will return the current document's content in plain text. The supported export file types are:
> - text _(default)_
> - html
> - raw _(warning: this is browser specific!)_

epiceditor/js/epiceditor.js
```diff
@@ -177,10 +177,14 @@
...
-      // Make sure to look for TWO spaces and replace with a space and &nbsp;
+      
+      // Make sure to there aren't two spaces in a row (replace one with &nbsp;)
       // If you find and replace every space with a &nbsp; text will not wrap.
       // Hence the name (Non-Breaking-SPace).
-      content = content.replace(/\s\s/g, ' &nbsp;')
+
+      content = content.replace(/<br>\s/g, '<br>&nbsp;')
+      content = content.replace(/\s\s\s/g, '&nbsp; &nbsp;')
+      content = content.replace(/\s\s/g, '&nbsp; ')
```
OscarGodson, line 187: :sparkles: 

Although ugly, this seems to have worked beautifully!

epiceditor/js/epiceditor.js
```diff
@@ -1518,17 +1541,28 @@
...
+      if (data !== undefined) {
+        delete data.content;
+      }
+      return data;
     }
     else {
-      return files;
+      for (file in data) {
+        if (data.hasOwnProperty(file)) {
+          delete data[file].content;
```
OscarGodson: Since you actually got the editor's contents to work correctly (which is amazeballs! :fireworks:), why not just keep `.content` in there, but here convert it to a friendly format like (instead of the `delete data[file].content` line):

``` js
data[file].content = self.exportFile(file);
```

I dont believe that actually replaces anything in localStorage, right? I looked back at some more code and I realized `.content` was often used for things like:

``` js
var documentId = '123-45-67890'; // Set by the server on page load or 
editor.on('update', function (file) {
  $.ajax({
    type: 'POST',
    data: {
      id: documentId
      content: file.content
    }
  });
});
```

So, if you dont have any strong objections to removing `.content` in there now that it _seems_ fixed in all my tests, I'd love to keep it in there or else this update is going to be rough for some projects. They'll have to refactor all their event handlers and code a lot because the filename isn't passed back `getFiles(foo)`. You'd have to manually figure out what file is the currently open file.

test/test.getFiles.js
```diff
@@ -44,7 +44,9 @@ describe('.getFiles([name])', function () {
     expect(fileCount).to.be(3);
   });
 
-  it('should return the correct file when the name is specified', function () {
-    expect(editor.getFiles(fooFile).content).to.be('foo');
+  it('should return just one file when the name is specified', function () {
+    var file = editor.getFiles(fooFile);
+    expect(file).not.to.be(undefined);
+    expect(file).to.have.property("modified");
```
OscarGodson: maybe check for `created` as well since you're checking for modified. And, if you add back content, I guess check for `content` too?

OscarGodson:
Alright made some comments and I played with it a lot. I'm pretty damn excited this is actually fixed.

I noticed something in IE10 and 9 while playing with this (and don't worry about fixing it in this PR because it happens in master and develop too). Type in a bunch of gibberish like:

```
flkjds lfjs flkdj fls jlf kdjflskj fdlksj flks jflkdj lfks jlkd jflsk fjlskd jflsdk jflsk jflksjf lkdsjf slfsdj flksjf lskjf lksdjf lksdj flks jflks jflks jflksdj flksdj flksdjf lksdjf lkdj flksdjf lkdjf lksj fkdj lskfj ldsf
```

Take note of how it's wrapping. Now, refresh. Notice that the wrapping is different. The actual data in localStorage appears to be correct at first glance, so not entirely sure whats up.

README.md
OscarGodson: Oh, and if you do this, you'll need to update the switch case in exportFile so that there's a case for raw that returns un-sanitized browser text.

OscarGodson:
Oh, and FYI, when I refreshed the page and everything looked the same with this patch, this was me:

epiceditor/js/epiceditor.js
Gankra: I was inclined to make it be "just" metadata in case in the future we want to split metadata/data into two separate stores to avoid parsing the contents of every single file just to get the timestamp of one. That said, I haven't noticed any substantial performance hits from it yet, so I'm fine with keeping it there if you want.

README.md
Gankra: Sounds great. Yeah I kinda gotta myself mixed up here.

epiceditor/js/epiceditor.js
Gankra: As a compromise, how about adding a flag to getFiles (excludeContent), so that you can make it only give metadata, but by default it does give you the content in "text" format.

Gankra:
Alright, how's that?

epiceditor/js/epiceditor.js
OscarGodson: Thats a good point. Whatever you think, it'd have to be like `getFiles([name], [excludeContent])`. Maybe it'd make more sense, like bash, zsh, etc, to have a getFileList() instead which just lists all files and meta data?

PR #240, opened 2013-06-11
Pasting HTML fix

Gankra:
Tested to work in:
- Firefox Aurora 23 (should work in Firefox 22, which comes out this month-- does _not_ work in FF 21, but will "gracefully" fall back to the old broken behaviour).
- Chrome latest stable
- Safari latest stable
- IE 10

No unit tests, since the clipboard is _very_ protected by modern browsers. Being able to read/write the clipboard automatically would be a severe vulnerability. As such, the clipboard is only exposed when the user triggers a copy/cut/paste via the browser's own chrome, and only in the actual event is the data exposed.

Should allow copy/paste of arbitrary text/html safely and "intuitively", while still keeping selections/undo-history working correctly.


Gankra:
Due to the FF 22 requirement, this definitely shouldn't be merged into any release branch.

Edit: (until it's released).

OscarGodson:
Testing might be hard. Don't have to test the insides since it'll break, but maybe you trigger a keyboard press and see if that function fires. You can also use jQuery to trigger "paste" on an element and make sure when the paste event happens the function runs. You could put it in a try/catch so the errors don't bubble during the tests too.

Gankra:
I suppose I could create a completely fake paste, but that would only address the "standards" path. I'm not sure if I can fake what IE does.

Gankra:
Yeah there's no way to tell the browser to run a fake paste, as far as I can tell. What's this about jquery? Is it actually available in the unit tests?

OscarGodson:
jQuery is required by Mocha, so should be. But I was thinking a simple test that when a paste event is fired on the pastable elements it fires the function:

http://jsbin.com/ikizez/1/edit

*I say simple, but it could very well be very hard. Just let me know and we could skip it for now, but if you can't test the jQuery way of triggering paste, I want to at least give it a shot.

Gankra:
Yeah I can add a trivial test like that, but it sure isn't going to really test anything meaningful.

Gankra:
http://jsbin.com/ivejed/1/edit

Expected: Paste?!  
Result: Paste!  
Conclusion: neither jQuery, nor any javascript, can trigger the (edit: native) paste event, as it is highly protected

So, there's no way to automatically test this functionality, unfortunately.

OscarGodson:
Oh well, no tests!

OscarGodson:
I'm thinking, if this works perfectly, we should close up the few PRs in the queue right now and release 0.2.2. This is a pretty big bug fix (yay!)

Gankra:
I will re-emphasize that this will not fix firefox until June 24, when FF 22 is released. But yes otherwise it seems to be working great.

OscarGodson:
What will happen if it's released before? Will it fail gracefully? Asking because everyone wont be upgraded on June 24th anyway, so this should fail gracefully.

Gankra:
Yep, if the "standard" FF/Chrome way, or the IE way are not available, it falls back to the old broken behaviour. I have confirmed this for FF 21.

Edit: numbers

OscarGodson:
So far this is working great. Very odd thing tho. I'm testing on Firefox 23 (nightly) and pasting actually works fine _without_ this code. Oh well, but wtf. It's like they're stripping HTML pastes by default.

OscarGodson:
Ha, I like the "do you want this web page to allow access to your clipboard data?" in IE.

Issue #130, opened 2012-06-10
Make docs readable on mobile


OscarGodson:
Right now the sidebar overlaps the right column when you zoom in on the right column making it almost impossible to read.

Gankra:
This is resolved, I believe.

OscarGodson:
Yep!

PR #241, opened 2013-06-11
Ticket #130 - docs are now minimally responsive

Gankra:
toc is hidden when it won't fit (very common pattern), and contents shrink-to-fit.

Far from "mobile first", as it were, but it's tolerable on a landscape phone. The CSS of the docs made me very sad inside (oh, oh the px...), but I'm not going to refactor the whole thing for my own satisfaction.

At worst, now you can keep the docs in a half-size window.

### DIFF
--- a/docs/css/main.css
+++ b/docs/css/main.css
@@ -205,3 +205,19 @@ p.btn-small {
 p.btn-small a {
   font-size:14px;
 }
+@media (max-width: 1050px){
+  #toc{
+   display: none;
+  }
+  #wrapper{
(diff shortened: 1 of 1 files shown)


OscarGodson:
A big reason it was px before was that when epiceditor was first released it didn't work very well in fluid layouts so we didn't want to make it that way. Now it's fluid left/right and also you can call `reflow()` manually. But yeah, not built well for this :( I'm just going to merge this as is. I'll play with it some tonight too.

PR #243, opened 2013-06-11
Focus and blur

Gankra:
Events now fire when editor is focused/blurred


OscarGodson:
Should fire on previewer too. Point is to allow you to know when a user is focused on EpicEditor and when they leave. Switching between editor and previewer shouldn't fire new focus or blurs tho since EpicEditor hasn't been blurred or focused.

Gankra:
I'm not sure focus is a meaningful sentiment to an uneditable document.

Gankra:
After a quick test, and look at docs, it seems that onfocus/onblur does not fire for non-editable elements (like iframes, documents, and divs). So do you want to cobble together a fake event based on mouse clicks or something crazy, or respect the standard meaning?

OscarGodson:
> I'm not sure focus is a meaningful sentiment to an uneditable document.

There's use cases for knowing when the user is using EpicEditor or not. A fairly simple, and not too uncommon use case, is while the user is in EpicEditor you want certain key events. Maybe a "save to dropbox" when you do {key command}. However, you want to release that handler when the editor is not in focus. Or, maybe you want to remap over cmd+o while in EpicEditor to run editor.open() (which works in preview), but when you blur you want to undo that mapping.

The issue you're having is why I haven't done this yet. The wrapping iframe focus/blur events would be super easy, but events like clicks "fall through" to the target element. The best idea I had come up with is when you clicked into it the editor or previewer it was focused. When you clicked on any element outside of EpicEditor it was blurred.

Gankra:
Hmm... I accept the use case, but I'm not sure I'm a fan of using standard terminology but subverting its meaning. Especially since editor.focus() actually is a proper focus on the editor.

OscarGodson:
Editor.focus() will focus on the editor OR the previewer :)

Gankra:
Okay. I'll go back to the drawing board on this one (might be a while on this one, bigger issues have cropped up for me).

PR #251, opened 2013-06-19
showButtons config proposal

Gankra:
Should probably write tests for this or something... but here's what the core changes are.


Gankra:
To be clear, this isn't "merge this in now!" but simply here for discussion of the change.

Gankra:
And there's the tests for this. Should be good now, assuming you like the config.

OscarGodson:
Sorry, been busy the last couple weeks. Is there a reason you made this a different option and not just part of `button` but like `bar` so:

``` js
var opts = {
  button: {
    fullscreen: false,
    bar: "show"
  }
}
```

If you'd like it to be a separate option, what about like `buttonBar`? `showButtons: "hide"` seems funny sounding :)

Gankra:
That would work too.

Gankra:
Proposed change implemented, should be good to merge.

OscarGodson:
This test is failing in Chrome:

51e02bc4e4b0e89eb245c614

Noticed if you do this in the console on the docs site:

``` js
editor.settings.button.bar = 'show';
```

The button bar never shows up. Probably why the test is failing?

Gankra:
This isn't reproducing on my machine. Maybe a cache thing?

OscarGodson:
The other options work and in Chrome I have caching turned off as well as I went to a new fake domain to test this via my hosts file. Still doesn't work. So, in the main docs' index, when you open the console and paste the code above, the button bar pops up from a display none?

OscarGodson:
From your code I don't see how flipping it on would turn it on. I see how hide would (you display none it), but not show as far as I can tell. Setting the setting programmatically isn't expected to work, but my guess is it's related to why the test is failing.

Gankra:
All my tests pass, and changing the defaults in main.js makes the docs page work as expected.

How 'show' works:
show->true @ 

```
    if (self.settings.button.bar === "show") {
      self.settings.button.bar = true;
    }
```

doesn't ever get hidden @

```
    if (self.settings.button.bar !== true) {
      utilBar.style.display = 'none';
    }
```

doesn't ever get hidden @ 

```
    function utilBarHandler(e) {
      if (self.settings.button.bar !== "auto") {
        return;
      }
```

In other words I basically prevent anything from ever hiding it.

OscarGodson:
> I basically prevent anything from ever hiding it.

It's already hidden on page load unless something is weird or custom with your docs because by default it's `auto` and then hides. Since you never make it display block it'll never re-appear. I'll take another look tonight tho.

Gankra:
?

The only code that hides it are the two places that I applied a diff to.

How are you applying the setting?

I may be misunderstanding how the settings are designed to be (re?)applied.

Gankra:
I just rebased my work on top of develop and it still seems to be working as expected (by me, which may be incorrect).

Gankra:
Just noticed there's a utilBar and utilBtns variable that seem to track the exact same value... dafuq?

OscarGodson:
So, if I put a debugger on L88 of test.options.js:

Gankra:
Broken image?

OscarGodson:
How about this? If not I can upload to imgur or something when I get to work

https://github-camo.global.ssl.fastly.net/328399751c234d960da0b6651f99daa82e883b94/687474703a2f2f6d6f6e6f736e61702e636f6d2f696d6167652f6a74574f6f38426366305872776471536374735145555573462e706e67

Gankra:
"Not found"

Could you push up the branch you're experiencing this trouble on somewhere? I'd like to see if our sources have diverged somehow.

Also I guess I have to get the offensively obvious question out of the way: did you do a new jake build after merging?

OscarGodson:
Just merged your branch into develop. There was a conflict in the min file (obviously), so I had to do a Jake build.

Gankra:
Dang, then we should have the same source... the hell...

Gankra:
shot

Gankra:
Look, I'm going to say it. 

We need to fight to the death.

It's the only fix I can think of.

Gankra:
143gankwoaaaaaaaaaaaaaaah

OscarGodson:
LOLOLOL

OK OK. WTF.

What version of chrome are you on? What about just changing the check for the `epiceditor-utilbar` instead?

I'm on: 28.0.1500.71

Gankra:
Until five seconds ago: 27.0.1453.116
Restarted browser and tried again just now on: 28.0.1500.71

Gankra:
Is this reproducing on Firefox or anything else?
Also can you push/host your branch to somewhere?

OscarGodson:
Does NOT happen in Safari and Firefox.
DOES happen in IE9, 10, 11, and Chrome.

Gankra:
Our two branches are identical (did a git diff), and I can't reproduce this bug on any browser in OSX or Ubuntu (literally just downloaded chrome on ubuntu just to check). Out of office, so I can't test a windows VM, but I literally can't see how it would _not_ work; especially the unit test.

This is insanity.

OscarGodson:
And, considering IE9 10 are on one machine, 11 on another, and chrome on another, it can't be caching or anything. I think you should just change the test to check for the button bar, not all the buttons. To be honest, it could just be a jQuery bug or some race condition?

Gankra:
Yes! haha! Got it running on my television/gaming pc so that I could reproduce it, and that fixes it!

OscarGodson:
:thumbsup:

Gankra:
That should do it!

OscarGodson:
ANNNNNNNND merged. Thank god :)

PR #257, opened 2013-06-24
ios scrolling

Gankra:
Moved everything into an internal <div> so that apple's "touch" scrolling can be enabled.

Seems to be working fine so far. Had to refactor some of your tests because they modified the body directly, but it should be good now.

In the ios simulator an anchor scroll test seems to fail, but that's fairly minimal as far as I'm concerned.

I've found that scrolling inside EE is rather... unreliable in the simulator. Oddly, it always works as expected in my barebones test of an iframe with the right stuff in it. Not sure what's up with that. If you have an iPad/iPhone I'd love to get your feedback on if the scrolling is now working on an actual device.

Consider this a "progress check-in", there's obviously more work to do.


test/test.load.js
```diff
@@ -92,7 +92,7 @@ describe(".load([callback])", function () {
       previewer = editor.getElement('previewer');
       editor.importFile(id, '<a href="#test">foo</a>\n\nfoo\n\nbar\n\n<h1 id="test">Test</h1>\n\nhello\n\nworld\n\nblah\n\nblah');
       editor.element.style.height = '10px';
-      editor.reflow().preview();
+      editor.preview();
```
Gankra, line 95: For some reason this reflow was causing chrome to return innerText of the div to be empty. Seemed fine to remove?

epiceditor/js/epiceditor.js
```diff
@@ -606,11 +607,18 @@
     self.previewerIframe.style.position = 'absolute';
 
     // Now grab the editor and previewer for later use
-    self.editor = self.editorIframeDocument.body;
+    self.editor = self.editorIframeDocument.getElementById("epiceditor-editor");
     self.previewer = self.previewerIframeDocument.getElementById('epiceditor-preview');
    
     self.editor.contentEditable = true;
- 
+    self.editor.style.height = "100%";
```
OscarGodson: There's an `_applyStyles` function you may want to use instead. Would be a little cleaner.

OscarGodson:
Sweet, I will have to try it out!

Gankra:
Eurgh... this is just... disgusting. I feel like my simulator is really borked or ios' safari is The Worst.

Scrolling with padding overflows outside of view, selection is screwy as hell. Nothing makes any sense.

Why Apple, why?

Gankra:
Found a _way_ better solution! Killing this one.

OscarGodson:
haha, nice, checking that one out now.

Issue #27, opened 2012-02-24
iOS Support


OscarGodson:
We need to make the editor work on mobile devices, but for 1.0 just iOS5 for now. If more, awesome. I'm sure it'd mostly be CSS with some minor JS checks for touch support.

Maybe even on focus the editor fills the phone so it's just the editor and keyboard.

LucaRosaldi:
That would be great. Don't forget to display the "preview" and "fullscreen" options which are on hover, though. When I first heard of EE, I opened the demo page with an iPad, I tried the editor and thought "so what?", before having a quick glimpse of the two icons in the right bottom corner and realizing there were other options.

OscarGodson:
@lucasrosaldi yep! The editor should have a different look for iOS altogether and will need to work with touch events. Hopefully I can get to this in the 0.1.3 release.

fuddl:
Should be renamed to 'Mobile support' since there is not only iOS any more. 

In Chrome for Android its a little hard to use right now.

OscarGodson:
Android would be a different ticket like how there was Safari, IE9, etc tickets. "mobile" support is too broad. If you want, go ahead and make an android ticket tho :) I want to add support eventually for sure, but the main issue is I don't have an android device.

fuddl:
ok, I thought this issue is about _how should things work on touch devices_ which would affect iOS and Android equally. 

> Maybe even on focus the editor fills the phone so it's just the editor and keyboard.

I would like that feature and I offer review it on Chrome for Android once there's a pull request.

OscarGodson:
We want the same experience like we want the same experience across desktop
browsers, but the code to achieve it it a quite a bit different for
different mobile devices. iOS will probably just be the first to implement
the mobile UX because that's what I have. Android would come next.

joemccann:
Status on this?

OscarGodson:
_Sigh_, soon. I had 2 core contributors but they've gotten busy with other projects/life so it's just me so progress has slowed down but I do generally work on this project every other day. It's set or 0.3.0 which does mean very soon since I'm close to the [0.2.1](https://github.com/OscarGodson/EpicEditor/issues?labels=0.2.1&page=1&state=open) release then `0.2.2` will probably be some even smaller fixes and me finishing up IE8 support. Then onto 0.3.0.

joemccann:
Nice one.  I'm probably going to drop this in as a replacement for Ace Editor on http://dillinger.io

[...]

PR #258, opened 2013-06-25
Ticket #27 - simple ios scrolling fix

Gankra:
Oh man this is so much better. Doesn't change anything important. Just adds some extra css to the wrapper iframe.

Also doesn't have _any_ of the stupid issues I was having before. Seems to work great, and doesn't effect anything else as far as I can tell.


OscarGodson:
haha. I actually didn't have multiple iframes before because it was going to be a pain. I ended up adding it because of the growing number of bugs between preview, base, and editor. I'm glad it worked out this time :)

Are these styles things that _need_ to be there for iOS tho? If so you should probably set them in the JS because these are supposed to be themes and therefore easy to modify without having it totally break or lose support for mobile.

Gankra:
Alright, good call. I've also noticed this line here: https://github.com/OscarGodson/EpicEditor/blob/develop/src/editor.js#L615

sets the height of the body on the outer-most iframe, but nothing ever seems to update it. This was fine before since the iframe was basically exploding out of the wrapper div (go ahead, inspect element, it has a computed height of "0px"!). However, now that I'm making the wrapper div actually constrict the bounds of the iframe, the fact that the body has a fixed height seems to be breaking reflow. 

It seems like setting height:100% on this line instead seems to work fine. Any thoughts on this?

OscarGodson:
> sets the height of the body on the outer-most iframe, but nothing ever seems to update it. 

It never needed to be updated if I recall correctly because the only time it changed was on fullscreen open/close which it is updated then.

Just make sure it works fine in firefox, especially and opening/closing fullscreen in native and non-native fullscreen works as expected.

Gankra:
Made the change, and everything seems to work right in firefox and all the other browsers I've tried.

OscarGodson:
Cool, i'll check it out tomorrow. Really busy tonight but should be able to tomorrow. Do you want it in this next minor release or should we hold off for 0.3.0?

Gankra:
Whichever you prefer. I merge my stuff into my own build as it gets okayed, so all that matters to me is that it's production quality. :)

Gankra:
Got my hands on an ipad, and it's not cooperating much. I'm thinking just letting the bloody thing autogrow on ios might just be the right thing.

OscarGodson:
So, what do you wanna do with this PR? Should I review this? I don't have an iPhone so, I'll test it in desktop Safari just to make sure there are no regressions, but I'll take your word if it works well and you wanna merge it :)

Gankra:
So after much work and testing, I'm leaning to the conclusion that inner scrolling areas is just _not_ mobile friendly. It's hard to browse, there's no indication that it _does_ scroll, and the platform is fairly bad at identifying what level of scroll you want to be interacting with. Therefore the only way to make the system ios (or android/wp/etc) friendly is to just enable autogrow (at least on that platform). Obviously autogrowing elements could break user layouts, so I'm inclined to just officially say "if you want to be mobile friendly, you should set autogrow:true", but keep autogrow:false as the default.

Currently hacking on the autogrow functionality.

Gankra:
So basically, I'm inclined to kill this PR at this point, but am willing to leave it up at the moment to hedge my bets.

OscarGodson:
@Gankro Feel free to merge this if you want. Since you're iffy on it I'll leave it up to you. It's a small change, so whatever you want. :)

Issue #259, opened 2013-06-28
Asset fingerprinting


pbougie:
Using the Rails asset pipeline with [asset fingerprinting](http://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark) is difficult with EpicEditor's current setup. I was able to include the CSS theme files by using the following configuration:

``` javascript
var opts = {
  basePath: '',
  theme: {
    base: "<%= asset_path 'epiceditor/epiceditor.css' %>",
    editor: "<%= asset_path 'epiceditor/epic-dark.css' %>",
    preview: "<%= asset_path 'epiceditor/github.css' %>",
  },
}
```

Filename extension is `.js.erb` so that it is first parsed by ERB and results in (fingerprints are based on file contents):

``` javascript
var opts = {
  basePath: '',
  theme: {
    base: "/assets/epiceditor/epiceditor-fe29664b99583cb091b9736a620b1aa2.css",
    editor: "/assets/epiceditor/epic-dark-e011369dac6ce0e33adfa283509a95bd.css",
    preview: "/assets/epiceditor/github-00404a923df019f174b8e68c117b75ef.css",
  },
}
```

The 3 images don't work with asset fingerprinting because the paths and filenames are hard-coded in the JavaScript. The solution I used was to change the button tags to <span> tags and embed the images as base64 encoded background images in the CSS file. Seems to work very well. Also reduces number of HTTP requests.

I'd like to submit a pull request to get this in the main project so I don't have to maintain this manually every time I update the library in my project.

Let me know what you think.

OscarGodson:
Im not sure how well basr64ing works in IE9-10, also it makes it more of a pain to swap out images when people want their own icon. However, we'd love a PR for whatever to make it work better with Rails. I'm not sure what you have in mind, but we're open to any ideas as long as it keeps it easy to setup and doesn't break previous versions too badly.

Let me know! This would be awesome to get fixed!

OscarGodson:
@pbougie Can this ticket be closed or is there some more CSS / theme work that has to be done?

OscarGodson:
Can you reopen and update the description of the ticket if there's still work to be done tho?

pbougie:
No, these changes make it much easier to work with the Rails asset pipeline or similar. No source files need changing anymore which was the point. Thanks.

PR #260, opened 2013-06-28
Issue #259 - Embed images in CSS so asset fingerprinting works.

pbougie:
Moves images into base64-encoded CSS to fix problems when using the Rails asset pipeline and asset fingerprinting. Changing an image is almost as easy as it was: instead of changing the image in the filesystem, you simply base64 encode it and paste it into the CSS. Use an encoder such as this one: http://webcodertools.com/imagetobase64converter


OscarGodson:
@Gankro what do you think about this?

Gankra:
Base64 embedding in CSS:
- more portable
- easy for users to clobber/replace
- less HTTP reqs
- well supported as far as I can tell
- better versionable?

Seems like a good call to me. My only issues are implementation specific.

pbougie:
> Since this is all in an iframe, its not clear to me that you need anything more than .epiceditor-fullscreen-foo as the selectors here.

I never touched any of the selectors. Any refactoring of selectors should probably be done elsewhere.

> Semantically, should these not be <button>?

Agreed. I made the changes.

> Not a huge fan of useless whitespace edits, but that's just me.

I mistakenly let those go through. I've reverted the whitespace edits. However, trailing whitespace should be removed project-wide but should be done in another commit.

OscarGodson:
@pbougie thanks. I'll take a look at this tonight or something. Did you check IE9-10?

pbougie:
Works well in IE9-10. Is there somewhere where you list browser compatibility for EpicEditor?

OscarGodson:
Thanks @Gankro! The code all looks good to me too. I'm going to test it locally real quick then merge it if it looks good.

OscarGodson:
IE9: 
screen shot 2013-07-06 at 1 11 24 pm

OscarGodson:
Buttons are still buttons and base64 doesnt seem to be working.

pbougie:
@OscarGodson Works fine here. Looks like the new styles are missing. Did you force reload the CSS?

OscarGodson:
I can try again later. Works fine in IE10 mode tho which means the caching shouldn't be a problem.

OscarGodson:
@pbougie Deleted everything (cookies, cache, etc) and finally it worked. Hopefully this isn't an issue when upgrading for people. 

Thanks again for the contribution :)

Issue #268, opened 2013-07-15
Chrome and IE9-10 replacing all spaces with &nbsp;


Gankra:
I don't know what happened, but suddenly every instance of EpicEditor (master, develop, my own fork) I look at now has this bug:
- Type in some text into epic editor
- Refresh
- Inspect DOM; All spaces are now `&nbsp;`

This really screws the word-wrapping, making the content hideous, to say the least.

@OscarGodson are you seeing this too?

Chrome 28.0.1500.71
OSX 10.8.4

Not reproducing in Safari/FF.

Really hoping my Chrome install is broken or something.

Gankra:
Just reproduced in a brand-new install of Chrome 28.0.1500.72 m in a basically clean Windows 7 Enterpise SP1 VM.

Fuck.

Gankra:
Did some quick hacking, this seems to work:

```
@@ -168,7 +168,8 @@
     // If you want to know why we check for typeof string, see comment
     // in the _getText function
     if (typeof document.body.innerText == 'string') {
-      content = content.replace(/ /g, '\u00a0');
+      content = content.replace(/\s\s/g, '\u00a0 ');
+      content = content.replace(/\s\s\s/g, '\u00a0 \u00a0');
       el.innerText = content;
     }
     else {
```

Gankra:
Hmm, no that screws up \n\s, working on it...

Gankra:
```
@@ -168,7 +168,10 @@
     // If you want to know why we check for typeof string, see comment
     // in the _getText function
     if (typeof document.body.innerText == 'string') {
-      content = content.replace(/ /g, '\u00a0');
+      content = content.replace(/\n /g, '\n\u00a0');
+      content = content.replace(/[ ][ ][ ]/g, '\u00a0 \u00a0');
+      content = content.replace(/[ ][ ]/g, '\u00a0 ');
       el.innerText = content;
     }
     else {
```

Seems to work well. [ ] is just a space, formatted so the linter doesn't freak out.

OscarGodson:
Problem with this is the damn minifier stripping those `\u00a0` out. Do you know of another good, trustworthy linter? What happens to Chrome if you make it use Firefox's text formatting (the `else` block)? Does it blow up there too?

I trust that you see this issue and this fixes it. I've been having issues with this chunk of code forever. It's probably time to write tests for it. If you dont want to for this ticket thats fine, I can handle it in a new ticket, but just let me know.

[...]

PR #269, opened 2013-07-15
Ticket #268 - potential fix to whitespace issues

Gankra:
Here's the proposed fix. I've tested against all our target browsers (personal eyeball tests and unit tests), and have it staged for production in my deployment as a hotfix.

I will not be writing unit tests, as I'm currently a bit busy, but I have posted recommendations for unit tests you may want to write in #268, should you want to start testing this behaviour.


Gankra:
What are the odds of getting this merged and released by wednesday? Evidently we're under a bit more pressure than I thought to ship _something_ for demo purposes. No one here wants to ship a tweaked fork of a third-party lib, and this is a pretty major bug if it's repro-ing on other machines.

OscarGodson:
Yeah, I was able to repro it too. I'm at OSCON today. I'll test it at lunch!

OscarGodson:
@Gankro If anything, this makes everything simpler and works more consistently across browsers. There's one odd thing in IE9 and 10 tho. Clear localstorage and start fresh. Type some text (notice the big gap above the text). Now refresh. Text moves closer to top of editor.

OscarGodson:
Oh, and this is merged by the way

Gankra:
Yeah that seems to be a thing with how IE treats divs vs brs. I think it does new lines with divs by default, whereas we use brs.

PR #287, opened 2013-09-12
Use _applyStyles instead of _applyAttrs when modifying the textarea

archseer:
Fixes a bug where the textarea style was set to "[object Object]" (a textual representation of the JSON) instead of the actual styles.


OscarGodson:
Rad, thanks! That's odd it was ever like that. I'd merge now but on a phone.

archseer:
Yeah, I was trying to integrate it instead of Pagedown and it was pretty bizarre seeing `style="[object Object]"` on the textarea

OscarGodson:
Merged. Thanks!

PR #296, opened 2013-10-12
Made callbacks optionnal

vinz243:
Made callbacks optional thus it will avoid some errors or bugs

### DIFF
--- a/src/editor.js
+++ b/src/editor.js
@@ -811,7 +811,9 @@
 
         self.emit('fullscreenenter');
 
-        callback.call(self);
+        
+        if(callback)
+         callback.call(self);
       }, wait);
 
     };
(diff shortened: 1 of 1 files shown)


OscarGodson:
Thanks! A few things:
1. Our style is to do `callback = callback || function () {}` at the beginning of the functions. Could you do that instead, or, at the least add curlys around those. We always add curlys in our codebase.
2. Do all tests pass? While this is not required, it'd rock if you added simple test cases for this (if callback is omitted, an error is NOT thrown)
3. Whenever you're done with 1 and 2 please build (jake build to build manually or jake watch to build while editing)

zethussuen:
Seems @vinz243 has removed his original PR branch. I can take a look into refactoring the code and adding tests, unless he's still interested?

PR #310, opened 2014-03-10
This is document fix. Try-it unload button css style height fix.

sagarjunnarkar:
When we unload after Try-it, the textarea disappears but the height
is not removed. So Text below textarea should come up after unloading.


OscarGodson:
Thanks!

PR #321, opened 2014-04-25
typo: it's -> its

case:
EpicEditor's great, thanks Oscar!


OscarGodson:
You need to update the README.md file. Index.html is just a generated file from that and will get overwritten the next build. Steps are:
1. Update readme
2. Run `jake docs`
3. Commit readme and index.html changes

PR #322, opened 2014-04-25
Remove `text-align:left` from github.css

tiziano88:
GFM has support for specifying alignment in tables by using colons in the header row (see https://help.github.com/articles/github-flavored-markdown#tables). Currently github.css overrides this by forcing left alignment for no good reason.


OscarGodson:
Thanks!

PR #323, opened 2014-04-25
Fix whitespace in editor

tiziano88:
Currently, pasting external content with sequences of whitespace into
the editor results in the whitespaces being collapsed into a single one
(which is the default behaviour for HTML elements). Instead we really
want all the whitespaces to be preserved and displayed correctly.

Note that this only applies when pasting content, since pressing the
space bar on the keyboard will produce `&nbsp;` which are not collapsed
in any case.

See http://www.w3schools.com/cssref/pr_text_white-space.asp for
more details.

### DIFF
--- a/epiceditor/themes/editor/epic-dark.css
+++ b/epiceditor/themes/editor/epic-dark.css
@@ -10,4 +10,5 @@ body {
   line-height:1.35em;
   margin:0;
   padding:0;
+  white-space:pre-wrap;
 }
--- a/epiceditor/themes/editor/epic-light.css
+++ b/epiceditor/themes/editor/epic-light.css
@@ -9,4 +9,5 @@ body {
   line-height:1.35em;
   margin:0;
   padding:0;
+  white-space:pre-wrap;
 }


OscarGodson:
I don't think this actually "fixes" the problem, does it? If we want to preserve white space shouldn't we convert the white spaces to `&nbsp;`? The CSS fix also means it'll break for every theme not included since its not in the JS and that, on refresh, those changes would go away, right?

tiziano88:
Well, it fixes it in the sense that it shows all the white spaces in the editor. Whether those spaces are preserved or not in the rendered output is another matter, and it depends on other factors like whether the text is inside a code block or not. Think of it as a "helper" for the user, in case they care about seeing all the whitespace while editing; in a way it is similar to the decision of having monospace font in the editor: "just in case" you are writing code or something for which alignment is important, but it doesn't mean that the output will all be monospace font.

OscarGodson:
I'd need to play with it and see the output but at the very least this should act the same across all themes as this isn't as much a styling thing as it is a functionality thing and so it should go in the JS where we set global styles.

@dadambickford, thoughts?

PR #325, opened 2014-05-09
Fix broken GitHub ribbon image src.

ygzgxyz:
It seems that `a248.e.akamai.net/camo.github.com` does not works now. So I fixed it to `camo.githubusercontent.com`. Copied new url from https://github.com/blog/273-github-ribbons. Changed only host, not path so theme of ribbon may be same with previous one.

### DIFF
--- a/docs/footer.html
+++ b/docs/footer.html
@@ -1,7 +1,7 @@
   <div id="avatars">
     
   </div>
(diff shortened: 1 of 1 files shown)


OscarGodson:
Thanks!

PR #327, opened 2014-05-13
AMD compatible

koorgoo:
I use Require.js to load modules. A little change is required to make EpicEditor compatible with AMD.

### DIFF
--- a/src/editor.js
+++ b/src/editor.js
@@ -1891,5 +1891,10 @@
   // Used to store information to be shared across editors
   EpicEditor._data = {};
 
+  if (typeof define === 'function' && define.amd) {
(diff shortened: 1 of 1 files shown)


OscarGodson:
Seems good to me. Couple things:
1. Are you comfortable writing a small test for this? Just make it so `define` is a function and `define.amd` is truthy and make sure define is called and returns EpicEditor?
2. Can you build the files? Just `jake build` https://github.com/OscarGodson/EpicEditor/wiki/Contributing

If not I'm not totally sure when I can get to it, but I can try later this week maybe.

koorgoo:
Nice. Will do after work. Thanks!

PR #329, opened 2014-05-17
Add property file name to default schema

calvernaz:
Add file name to default schema, useful for exporting purposes.
- Useful to export files and save on disk

### DIFF
--- a/src/editor.js
+++ b/src/editor.js
@@ -456,9 +456,10 @@
     // Setup local storage of files
     self._defaultFileSchema = function () {
       return {
-        content: self.settings.file.defaultContent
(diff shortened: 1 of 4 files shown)


OscarGodson:
Looks good, curious tho, why doesn't using the key work since the key is the file name?

calvernaz:
I could augment the object on export, but:
-  this change fits better my use case, as I load the editor for each file, and export like this.
  
  var content = this.editor.exportFile(null, 'json');
- cleaner, getting the content using the file name and augment the content (the exported result) using the file name, requires one more unnecessary step, and one unnecessary input like I described above

PR #332, opened 2014-06-26
Added support for bower

brunowego:
For better integration with bower is necessary this file.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,27 @@
+{
+  "name": "EpicEditor",
+  "version": "0.2.2",
+  "homepage": "http://epiceditor.com/",
+  "authors": [
+    "Oscar Godson <oscargodson@outlook.com>"
+  ],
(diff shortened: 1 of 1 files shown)


brunowego:
If merge, make a new tag. Thanks.

mrhanlon:
Thanks for this, +1. I was about to propose the same.

mrhanlon:
Looking at your bower.json, on line 10 there is an extra trailing comma that is invalid. Also, EpicEditor doesn't actually depend on jQuery, only the docs example does. I recommend to remove it from the dependencies.

zackyang000:
+1

brunowego:
Yes @mrhanlon, fixed now! Thanks!

brunowego:
Waiting for merge!

OscarGodson:
EpicEditor doesn't require jQuery. I'm going to go ahead and give you access to merge this yourself @brunobatista. I've never used Bower so I wouldn't even know if this is correct or not. But, jQuery is not a dep of EpicEditor so I would assume that could be removed.

brunowego:
New pull in #348

PR #379, opened 2015-11-18
Add very simple Debian packaging

massar:
This allows using Debian package management to install EpicEditor and also have other tools using EpicEditor depend on this package for easy installation (and thus avoiding having to embed Epic Editor in their packages independently).

on my sbuild box this builds fine:

```
$ sbuild -v -d jessie --arch-all
....
Build Architecture: amd64
Build-Space: 1100
Build-Time: 1
Distribution: jessie
Host Architecture: amd64
Install-Time: 18
Job: /home/jeroen/git/epiceditor_0.2.5.dsc
Lintian: pass
Machine Architecture: amd64
Package: epiceditor
Package-Time: 28
Source-Version: 0.2.5
Space: 1100
Status: successful
Version: 0.2.5
```

and produces:

```
-rw-rw-r-- 1 jeroen users    725 Nov 18 16:24 epiceditor_0.2.5.dsc
-rw-rw-r-- 1 jeroen users 272460 Nov 18 16:24 epiceditor_0.2.5.tar.xz
-rw-rw-r-- 1 jeroen users  39682 Nov 18 16:24 epiceditor_0.2.5_all.deb
-rw-rw-r-- 1 jeroen users    670 Nov 18 16:24 epiceditor_0.2.5_amd64.changes
```

Effectively it is almost a tarball of course with version details, thus extremely simple, but quite useful.

Next step would be to ask a friendly Debian person to use this to include it directly into Debian...


OscarGodson:
I can't really test this on my end but I'm totally open to letting you merge it. I'm curious tho what a Debian install is useful for since it's a web module?

massar:
EpicEditor is used by Trident (https://trident.li ; see presentations for screenshots etc) -- and for that we will be providing Debian packages, thus instead of including/embedding EpicEditor in those packages the right thing is to have a separate package so that any tool using EpicEditor can just depend on the epiceditor package and voila, they can serve /usr/share/epiceditor/js/epiceditor.min.js along with the CSS and presto.

Then also updates will automatically spread to all packages that depend on it instead of having to update everything that embeds/includes it.

(of course compatibility is something funny there, but that is why there is versioning and Debian has version requirements on packages etc)

OscarGodson:
@massar very cool! I went ahead and made you a contributor so you should be able to merge this yourself. Feel free to merge when you feel ready.

PR #383, opened 2015-12-20
Make epiceditor-previewer a CSS class

kousu:
This is with the goal of using the preview styles as the real site-wide styles in mind. Right now the CSS uses a div ID, meaning I can't just pluck the stylesheet and use it to style the posted markdown after I edit it.

Great editor by the way! It's refreshing to see clean APIs.


OscarGodson:
These are all within an iframe so these IDs shouldn't have any affect on your site. Or do you mean you want to use more `epiceditor-*` within the iframe?

kousu:
Hi Oscar,

The isolated iframe is a good idea  I'm not interested in the div, I want to use the stylesheet. As it is, I have to write something like <div id=content><div id= epiceditor-previewer> to get it to kick in on my site.

I should probably use a global stylesheet that sets the styles for <code> and whatnot globally, but for now reusing yours is the fastest way to make it look consistent.

On December 20, 2015 1:48:00 PM EST, Oscar Godson notifications@github.com wrote:

> These are all within an iframe so these IDs shouldn't have any affect
> on your site. Or do you mean you want to use more `epiceditor-*` within
> the iframe? 
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/OscarGodson/EpicEditor/pull/383#issuecomment-166146699

## 

Nick Guenther
4B Joint Stats/CS
University of Waterloo

OscarGodson:
oh i see, interesting use case. I'm fine with it. I made a collaborator so that you can merge this when you're ready. There's some things I would suggest running tho before you merge:
1. Update the docs that mention the ID stuff. You'll see in the docs towards the bottom the markup that EpicEditor uses.
2. Run the tests. The tests should pass but if not it's probably just because it's looking for a `#epiceditor-*` selector. Just change those to be your new selector.

OscarGodson:
P.S. There's docs on updating the docs and running the tests here: https://github.com/OscarGodson/EpicEditor/wiki/Contributing#updating-docs

kousu:
Ah! Thank you for the vote of confidence. It's merged now and I hope I didn't screw it up. Your CONTRIBUTING.md is super clear, by the way.

OscarGodson:
:+1:

PR #384, opened 2015-12-21
Add explicit installation instructions.

kousu:
Usage instructions were not enough. I had to guess my way through it. I stuck in a new Step 2 to help others stuck like I was. Check it and tell me if you think my writing is _too_ verbose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,26 @@ EpicEditor is easy to implement. Add the script and assets to your page, provide
 $ git clone git@github.com:OscarGodson/EpicEditor
 ```
 
-### Step 2: Create your container element
+### Step 2: Install
+
+Copy `EpicEditor/epiceditor/` onto your webserver at the same depth as the page you intend to use it in.
+
+```bash
+$ ls -l
+[user@site public_html/new_stuff/my_blog]$ ls -l
+total 8
+drwxr-xr-x  4 user www  4096 Dec 20 05:10 epiceditor
+-rwxr-xr-x  5 user www  4096 Dec 20 06:23 editor.html
+[user@site public_html/new_stuff/my_blog]$ ls -l epiceditor/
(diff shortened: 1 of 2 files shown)


OscarGodson:
Good idea! 

You can probably do without the `ls` stuff. The key point which you made is that `EpicEditor/epiceditor` contains the distribution files (`/epiceditor` should probably just be renamed eventually to `dist` or something of that nature). Maybe something like this?

> `EpicEditor/epiceditor/` contains the JS files in `js`and CSS in `themes`. Copy those into their respective folders on your webserver
> 
> Alternately you can install it globally, say in /static/lib/epiceditor, and set the basePath option to the same to point EpicEditor in the right direction.

kousu:
My guess was that epiceditor was supposed to be kept together in `epiceditor/`, which is how other js libraries like Lightbox work, so I decided not mention about "respective folders". If someone has that need they can figure it out for themselves.

I also decided to explicitly tell people where to install to, for simplicitly. I did this because the default of `basePath = "epiceditor"` means EpicEditor uses relative links to find its files, doesn't it? So you would have to have a different copy for every page you use it on.

OscarGodson:
I like the simplicity of how you have it now. EpicEditor uses relative links, yes, so `epiceditor/themes/base/epiceditor.css`, but you can modify any of that so `basePath` can become `/epiceditor` which should always be the same path no matter how deep or `/css/epiceditor.css` by altering the `settings.theme.base`.

EpicEditor tries to be as flexible as possible with putting it however you want. People use this on fully static sites and within one page apps without URLs and through things like asset pipelines. I never know how it's going to be used. The simplicity of your example is nice because if you need something more advanced you can just go and read more in the options list.

kousu:
Awesome. Merged. :green_apple:

PR #394, opened 2016-03-12
fixes #378, broken twitter-avatars

### LINKED ISSUE #378: Broken Avatars - Footer
In footer the four images are broken.
not

Code tries to access avatars from Twitter using `http://twitter.com/api/users/profile_image?screen_name`. But Twitter API doesn't serve this anymore.  So either we need to place avatar url directly in the code or use this way (http://stackoverflow.com/a/17110904/2820154) to get the avatars.

madpew:
the old public-api doesn't exist anymore, new api needs auth.
Links in the form of twitter.com/[username]/profile_picture do work.


massar:
As you're updating it, I would suggest updating the URL to https://twitter.com too...

Issue #402, opened 2016-12-06
Update website (itself, readme, or repo description)


0joshuaolson1:
epiceditor.com doesn't work. How about http://stacks.math.columbia.edu/js/EpicEditor/ ?

pmackay:
At least just remove the link from the top of the repo. It's been hijacked now I think..

JohnDarkhorse:
> At least just remove the link from the top of the repo. It's been hijacked now I think..

Confirmed.  The site's displaying a popup asking you to "Before you continue, please download a <browser> extension to help safely navigate the web.  <Browser> extension will assist by changing your default search provider"

PR #404, opened 2020-03-11
add deprecation warning to README

zethussuen:


zethussuen:
A response to quite an old issue #402

